PR #11, opened 2017-12-14
Merged Deric's change on chris2 branch and added addresses dropdown and Bet status text

chrisli30:


src/components/bodhi-dls/cardVoting.js
```diff
@@ -34,12 +33,35 @@ class CardVoting extends Component {
...
+          message="Oops, something went wrong"
+          description={result.error}
+          type="error"
+          closable={false}
+        />);
+      }
+    }
+    const alertContainer = <div className="alert-container">{alertElement}</div>;
+
+
```
dwalintukan: :space_invader:

src/containers/Topbar/Topbar.js
```diff
@@ -12,6 +12,23 @@ import { themeConfig } from '../../config';
 
 const FormItem = Form.Item;
 const { Header } = Layout;
+const addressMaxDisplayLength = 11;
```
dwalintukan: `ADDRESS_MAX_DISPLAY_LENGTH = 11;`? I think we should start standardizing the naming conventions.

src/containers/Topbar/Topbar.js
```diff
@@ -12,6 +12,23 @@ import { themeConfig } from '../../config';
...
+const KEY_ADD_ADDRESS_BTN = 'add_address';
+
+function shortenAddress(text, maxLength) {
```
dwalintukan: can this be a static function?

src/containers/Topbar/Topbar.js
```diff
@@ -12,6 +12,23 @@ import { themeConfig } from '../../config';
...
+  let ret = text;
+
+  const startLen = (maxLength - 3) / 2;
+  const endLen = (maxLength - 3) / 2;
+
+  console.log(`ret is ${ret}`);
+
+  if (ret.length > maxLength) {
+    ret = `${ret.substr(0, startLen)} ... ${ret.substr(ret.length - endLen)}`;
```
dwalintukan: the end part should be `${ret.substr(ret.length - endLen - 1)` right? since its based on 0 index?

src/containers/oracle.js
```diff
@@ -221,6 +223,9 @@ TopicPage.propTypes = {
   match: PropTypes.object,
   onBet: PropTypes.func,
   betResult: PropTypes.object,
+  walletAddrs: PropTypes.array,
+  walletAddrsIndex: PropTypes.number,
+
 };
 
 TopicPage.defaultProps = {
```
dwalintukan: oh this class should prob be renamed to `OraclePage` or `OracleDetails`. Forgot to change that sorry.

src/containers/Topbar/Topbar.js
chrisli30: The first parameter of substr is start index, including itself
Let's say we have a string of 20 "01234567890123456789", maxLength is 11
endLen = (11-3) /2 = 4;
ret.subStr= (20 - 4) = ret.subStr(16);
So the end part will be the 16th,17th,18th and 19th, which are expected.

PR #19, opened 2017-12-18
Removed toggle in topbar.js and container/sidebar

chrisli30:
Code cleanup


src/redux/app/actions.js
```diff
@@ -9,7 +9,6 @@ export function getView(width) {
 }
 
 const actions = {
-  COLLPSE_CHANGE: 'COLLPSE_CHANGE',
   COLLPSE_OPEN_DRAWER: 'COLLPSE_OPEN_DRAWER',
```
dwalintukan: Can we get rid of this one too? Seems like the drawer related logic is the deleted from the sidebar.

src/redux/app/reducer.js
```diff
@@ -36,8 +36,6 @@ export default function appReducer(state = initState, action) {
 
       return state.set('walletAddrs', result);
     }
-    case actions.COLLPSE_CHANGE:
-      return state.set('collapsed', !state.get('collapsed'));
     case actions.COLLPSE_OPEN_DRAWER:
```
dwalintukan: can we remove?

src/redux/app/actions.js
chrisli30: okay, will do a grep and remove it.

PR #31, opened 2017-12-19
Fixed a "can't get key of undefined problem" and added error handling logic to oracle page

chrisli30:


src/containers/oracle.js
```diff
@@ -134,6 +135,7 @@ class OraclePage extends React.Component {
               // Finalize oracle if current block has passed arbitrationEndBlock and threshold is not met
               // Since new oracle is guarranteed to spawn if total amount threshold is met we are not checking total BOT amount here
               if (blockCount > oracle.endBlock) {
+                console.log(`!! blockCount ${blockCount} is greater than endBlock ${endBlock}.`);
                 configName = 'FINALIZING';
```
chrisli30: console.log here for debugging purpose. Will clear them up later

src/containers/oracle.js
```diff
@@ -148,9 +150,14 @@ class OraclePage extends React.Component {
             configName = 'SETTING';
             break;
           default:
+            console.warn('Oracle exists but cant determine status. ');
+            oracle = undefined;
             break;
```
chrisli30: Basically, if we can't identify an oracle's status with above rules, we are gonna show 404 in the center of page, to which I haven't added style yet.

src/containers/oracle.js
```diff
@@ -201,16 +207,18 @@ class OraclePage extends React.Component {
...
+          console.log('onSetResult');
           onSetResult(contractAddress, selectedIndex, senderAddress);
         }, SUB_REQ_DELAY);
         break;
 
       case 'VOTING':
 
         /** The amount of voting needs to be approved by Bodhi_token * */
-        onApprove(oracle.topicAddress, amount, senderAddress);
+        onApprove(_.trimStart(oracle.address, '0x'), amount, senderAddress);
```
chrisli30: Every address's "0x" needs to be trimmed before send to api. I am open to suggestions.

dwalintukan: Already fixed in API/qweb3 layer. No need to trim any addresses of any sort of the `0x`.

src/containers/oracle.js
dwalintukan: 👍

src/containers/oracle.js
dwalintukan: 👍

PR #49, opened 2017-12-21
UI fixes for demo

chrisli30:


src/components/bodhi-dls/cardVoting.js
```diff
@@ -92,6 +92,9 @@ class CardVoting extends Component {
       // Disable the button if request went through
       confirmBtnDisabled = true;
       buttonText = 'Transaction posted';
+    } else if (skipToggle) {
+      confirmBtnDisabled = false;
+      buttonText = 'Finalize';
     } else if (showAmountInput) {
```
chrisli30: Hacky code here. Currently only Finalizing phase will set skipToggle to true so buttonText is hard-coded as "Finalize". The reason we need this is that Finalizing phase doesn't have any input or selection so it doesn't need toggle.

src/containers/Topbar/Topbar.js
```diff
@@ -201,9 +201,8 @@ class Topbar extends React.PureComponent {
         </Link>
       ) : (
         <Dropdown overlay={menu} placement="bottomRight">
-          <a className="ant-dropdown-link" href="#">
+          <a className="ant-dropdown-link" onClick={(evt) => { evt.preventDefault(); }}>
             {this.getSelectedAddressObject() ? shortenAddress(this.getSelectedAddressObject().address, ADDRESS_TEXT_MAX_LENGTH) : null}
-            {this.getSelectedAddressObject() && !_.isUndefined(this.getSelectedAddressObject().qtum) ? this.getSelectedAddressObject().qtum.toFixed(1) : null}
             <Icon type="down" />
```
chrisli30: Removed qtum amount from circled button cuz it's cluttered with address and hard to see.

PR #56, opened 2017-12-25
Plugged in /botbalance endpoint and amount BN hex string change from DB

chrisli30:


src/helpers/utility.js
```diff
@@ -136,3 +139,16 @@ export function randomString(size) {
...
+
+/**
+ * Convert a BigNumber to ES6 Int (2^53 max == 9 007 199 254 740 992) and divide it by 10^8
+ * If result number is too small (less than 0.01) we return 0
+ * @param  {[type]}
+ * @return {[type]}
+ */
+export function convertBNHexStrToQtum(input) {
+  const bigNumber = new BN(input, 16);
+  const value = bigNumber.toNumber() / BOTOSHI_TO_BOT;
```
dwalintukan: Do `bigNumber.div(BOTOSHI_TO_BOT).toNumber()`. This will make the calculation as a big number. 

Converting to a ES6 number first before the calculation defeats the purpose of having it as a bigNumber as you will get data loss (2^53 vs 2^256).

chrisli30: The reason I did it this way is because bn.js doesn't support decimal.....
Note: decimals are not supported in this library.

So if the end value is 1.2 it's 1 in your example.

by the way, I've tested it. The var in .div() needs to be a BN too.
const value = bigNumber.div(new BN(BOTOSHI_TO_BOT)).toNumber();

dwalintukan: how about:
```
const botoshi = new BN(BOTOSHI_TO_BOT);
const wholeNum = bigNumber.div(botoshi);
const remainder = bigNumber.mod(botoshi);
const numStr = (wholeNum.toString()).concat(".").concat(remainder.toString());
return parseFloat(numStr);
```

Not that elegant but it should work without losing any data.

chrisli30: I need number on UI side, are you concerned that numbers are still too big after broken into two parts?

2^53 == 9 007 199 254 740 992

Updated to this:
  const integer = bigNumber.div(botoshi).toNumber();
  const decimal = bigNumber.mod(botoshi).toNumber();
  const result = integer + (decimal / BOTOSHI_TO_BOT);

src/containers/topic.js
```diff
@@ -49,7 +50,34 @@ class TopicPage extends React.Component {
...
+
+  componentWillUnmount() {
+    this.props.onClearRequestReturn();
+    this.props.clearEditingToggled();
+  }
+
+  /** Withdraw button on click handler passed down to CardFinished */
+  onWithdrawClicked(obj) {
+    const senderAddress = this.getCurrentSenderAddress();
+    const contractAddress = 'fe99572f3f4fbd3ad266f2578726b24bd0583396';
```
dwalintukan: use `topic.address`

PR #70, opened 2017-12-28
Fixed Approve, CheckAllowance and SetResult/Vote lifecycle issue

chrisli30:


src/containers/oracle.js
```diff
@@ -396,26 +384,57 @@ class OraclePage extends React.Component {
...
+    const self = this;
 
-      this.props.onAllowance(selectedWalletAddress, this.state.oracle.topicAddress, selectedWalletAddress);
-    } catch (err) {
-      console.log(err.message);
+    // A function to repeat itself until this.state.isApproving is false
+    function startPollAllowance() {
+      if (self.state.isApproving) {
+        onAllowance(selectedWalletAddress, oracle.topicAddress, selectedWalletAddress);
+        setTimeout(startPollAllowance, ALLOWANCE_TIMER_INTERVAL);
```
dwalintukan: If you don't set this to a var, you can't cancel the timer when you leave the page. Or did you intentionally have it still running in the background when you leave the page? 

If so, this could be tricky say if you try to set result on 3 different Oracles. Then there would be 3 timers running in the background.

src/redux/topic/saga.js
```diff
@@ -92,9 +92,11 @@ export function* allowanceRequestHandler() {
 
       const result = yield call(request, `${bodhiapi}/allowance`, options);
 
+      const value = convertBNHexStrToQtum(result.remaining);
```
dwalintukan: nice change to have all the hex converted before being used by the UI 👍

src/containers/oracle.js
```diff
@@ -396,26 +384,57 @@ class OraclePage extends React.Component {
...
+  /**
+   * Determine next action based on returned allowance result
+   * 1. If allowance is zero, send approve request
+   * 2. If allowance is positive but less than voteAmount, send approve(0) to reset allowance
+   * 3. If allowance is greather than or equal to voteAmount and not isApproved, do action and reset isApproving and isApproved
+   * @param  {number} allowance value
+   * @return {}
+   */
   onAllowanceReturn(allowance) {
+    if (_.isUndefined(allowance) || _.isUndefined(this.state.config)) {
```
dwalintukan: after testing out the flow. it looks like `allowance` doesn't get reset to `undefined` after the setResult or vote. So onAllowanceReturn() is being called over and over even after the flow is done.

This doesn't affect setResult since you can only do it once. But, this can affect vote in this case:
- Place vote to 50 BOT
- approve() for 50
- allowance, wait, allowance etc
- approve accepted for 50
- vote() called (onAllowanceReturn.allowance still returning as 50)
- user wants to vote() again for 30
- onAllowanceReturn() called when /getblockcount or /listunspent returns. (anytime componentWillReceiveProps gets called)
- will call vote() because allowance has not been reset from 50
- the 2nd vote() for 30 fails because the approve() has not been done

 Need to reset `allowanceReturn`  in `componentWillReceiveProps` after the setResult or vote is done.

src/containers/oracle.js
chrisli30: I think you mistook setTimeout() for setInterval(). The former only set timer once and needs to call it self in execution block to repeat. When SetResult() or Approve(), isApproving is set to false, so timer is not set again in the last round.

src/containers/oracle.js
chrisli30: Good point. Also if allowance is not reset, if a new Vote comes in and calls onAllowance with the same value onAllowanceReturn will not be called, because componentWillReceiveProps is only triggered when there's any value change.

chrisli30: Verified your theory with e2e testing. It's so slow on testnet. I plan to switch to -regtest for development testing.

src/redux/topic/saga.js
chrisli30: 🙆‍♂️

src/containers/oracle.js
dwalintukan: oops forgot to remove this comment. sorry, ya I tested it out and there was only 1 timer.

PR #90, opened 2018-01-11
delete all unused files

henryzbw:


chrisli30:
themeSwitcher 
src/components/topbar/topbarSearch.js
needs to stay.

PR #91, opened 2018-01-12
Beta api changes

dwalintukan:
**DON'T MERGE** until corresponding API PR is merged: https://github.com/bodhiproject/bodhi-api/pull/39

### DIFF
--- a/src/config/routes.js
+++ b/src/config/routes.js
@@ -12,9 +12,10 @@ const Routes = {
   createTopic: `${bodhiapi}/createtopic`,
   bet: `${bodhiapi}/bet`,
   setResult: `${bodhiapi}/set-result`,
-  vote: `${bodhiapi}/vote`
+  vote: `${bodhiapi}/vote`,
   finalizeResult: `${bodhiapi}/finalize-result`,
   withdraw: `${bodhiapi}/withdraw`,
+  winnings: `${bodhiapi}/winnings`,
 };
 
 export default Routes;
--- a/src/config/routes.js
+++ b/src/config/routes.js
@@ -6,7 +6,15 @@ const Routes = {
   listUnspent: `${bodhiapi}/list-unspent`,
   getAccountAddress: `${bodhiapi}/get-account-address`,
   getBlockCount: `${bodhiapi}/get-block-count`,
+  approve: `${bodhiapi}/approve`,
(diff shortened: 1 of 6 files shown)


src/config/routes.js
```diff
@@ -0,0 +1,21 @@
...
+  createTopic: `${bodhiapi}/createtopic`,
+  bet: `${bodhiapi}/bet`,
+  setResult: `${bodhiapi}/set-result`,
+  vote: `${bodhiapi}/vote`,
+  finalizeResult: `${bodhiapi}/finalize-result`,
+  withdraw: `${bodhiapi}/withdraw`,
+  winnings: `${bodhiapi}/winnings`,
+};
+
+export default Routes;
```
chrisli30: This is neat!

src/redux/topic/saga.js
```diff
@@ -223,51 +222,21 @@ export function* calculateQtumWinningsRequestHandler() {
...
-
-      const result = yield call(request, `${bodhiapi}/botwinnings`, options);
-
-      const value = result ? convertBNHexStrToQtum(result['0']) : undefined;
+      const result = yield call(request, Routes.winnings, options);
+      let value;
+      if (result) {
+        const botWon = convertBNHexStrToQtum(result['0']);
+        const qtumWon = convertBNHexStrToQtum(result['1']);
+        value = [botWon, qtumWon];
```
chrisli30: You are doing action.qtumWon and action.botWon in reducer. Should this be value ={
botWon,
qtumWon
}
?

src/config/routes.js
dwalintukan: 👍

src/redux/topic/saga.js
dwalintukan: Sure. Changed to use an object instead. Can you test with yours? For the completed events, none of mine have any winning amounts > 0.

dwalintukan: Ah wait haha, none of the existing contracts on chain will be able to return anything since they dont have the `calculateWinnings` method. Will just return 0, 0.

Gonna deploy some more beta contracts for next round of beta testing.

chrisli30: Yeah, sure we can test when it's switched. Just need to remember this change.

PR #95, opened 2018-01-18
fix cancel should change to back after confirmation

henryzbw:
screen shot 2018-01-18 at 1 08 51 pm
After confirmation, cancel changes to back

### DIFF
--- a/src/components/bodhi/createTopic.js
+++ b/src/components/bodhi/createTopic.js
@@ -219,15 +219,10 @@ class CreateTopic extends React.Component {
               htmlType="submit"
               disabled={createReturn && createReturn.result}
             >Publish</Button>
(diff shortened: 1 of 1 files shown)


src/components/bodhi/createTopic.js
```diff
@@ -219,10 +219,15 @@ class CreateTopic extends React.Component {
...
+            {(createReturn && createReturn.result) ?
+              <Button
+                type="default"
+                onClick={this.onCancel}
+              >Back</Button> : <Button
+                type="default"
+                onClick={this.onCancel}
+              >Cancel</Button>
+            }
           </FormItem>
```
chrisli30: Since onClick action is the same, put status check inside <Button>

<Button> condition? Back: Cancel </Button>

PR #97, opened 2018-01-22
Beta changes

dwalintukan:


src/network/graphDataStruct.js
```diff
@@ -0,0 +1,86 @@
+import _ from 'lodash';
```
frankobe: wanna any lib handling import schema.js from grapqhql, so dunt need re-define here

src/network/graphParser.js
```diff
@@ -0,0 +1,56 @@
...
+      allOracles: this.parseOracle,
+      syncInfo: this.parseSyncInfo,
+    };
+    return PARSER_MAPPINGS[requestName];
+  }
+
+  static parseTopic(data) {
+    return data.map((entry) => ({
+      address: entry.address,
+      creatorAddress: entry.creatorAddress,
```
frankobe: removed

src/network/graphParser.js
```diff
@@ -0,0 +1,56 @@
...
+      name: entry.name,
+      options: entry.options,
+      bettingEndBlock: entry.blockNum,
```
frankobe: is this correct?

src/network/graphRequest.js
```diff
@@ -0,0 +1,106 @@
...
+      _.forEach(this.filters, (obj, index) => {
+        if (!_.isEmpty(filterStr)) {
+          filterStr = filterStr.concat(',');
+        }
+        const str = Object
+          .keys(obj)
+          .map((key) => {
+            if (isValidEnum(key, obj[key])) {
+              // Enums require values without quotes
+              return `${key}: ${obj[key]}`;
```
frankobe: Suggestion: all keys in grapghql query doesnt require quotes, only string value require it. So we should add quotes when value is string type

src/network/graphRequest.js
```diff
@@ -0,0 +1,106 @@
...
+            }
+            return `${key}: ${JSON.stringify(obj[key])}`;
+          })
+          .join(',');
+        filterStr = filterStr.concat(`{${str}}`);
+      });
+
+      query = query.concat(`(
```
frankobe: Suggestion: using string template

src/network/graphDataStruct.js
dwalintukan: Ya I was wondering about that too. Will look into it.

src/network/graphParser.js
dwalintukan: fixed

src/network/graphParser.js
dwalintukan: nope, removed. doesn't have that field anymore.

src/network/graphRequest.js
dwalintukan: done, good call 👍

src/network/graphRequest.js
dwalintukan: done

src/components/bodhi/createTopic.js
```diff
@@ -167,39 +176,69 @@ class CreateTopic extends React.Component {
...
-          >
-            {getFieldDecorator('bettingEndBlock', {
+
+          {this.createInputNumberField(
+            'bettingStartBlock',
+            'Betting Start Block',
+            {
               rules: [{
-                required: true, message: 'Please enter a future block number.',
+                required: true, message: 'Must be greater than or equal to current block number.',
```
frankobe: [open question] should we prevent the endblock number too close to current/start block num

dwalintukan: Maybe we can add a warning when we add the date/time picker. But don't think we should restrict the user since its an open platform and shouldn't impose rules that arent necessary.

PR #104, opened 2018-01-28
Time picker for block numbers

dwalintukan:
Add time picker instead of inputting block.
Adding logic to send unix timestamp to API instead of block number.


src/components/bodhi/createTopic.js
```diff
@@ -262,14 +161,298 @@ class CreateTopic extends React.Component {
...
+          bettingEndTime,
+          resultSettingStartTime,
+          resultSettingEndTime,
+        } = values;
+
+        this.props.onCreateTopic({
+          centralizedOracle,
+          name,
+          results,
+          bettingStartTime: bettingStartTime.utc().unix().toString(),
```
frankobe: toString() is not neccessary

src/helpers/utility.js
```diff
@@ -1,74 +1,13 @@
...
-    const minutes = Math.floor((temp %= 3600) / 60);
-    if (minutes) {
-      return `${minutes} minute${numberEnding(minutes)} ago`;
-    }
-    return 'a few seconds ago';
-  };
-  return getTime();
-}
-
+const AVG_BLOCK_TIME_SECONDS = 144.3489932885906;
```
frankobe: Is this the same value for testnet?

src/components/bodhi/createTopic.js
dwalintukan: On the UI -> qweb3 encoding side, I decided to pass in all and any UINTs as a string (non-hex). This solves a couple of issues:

1. A string representation of a decimal number handles any range of number that is limited by JS's limit of 2^53.

2. Not having the need to convert it to a hex - on qweb3 side, I am purposefully forcing the user to use the `0x` prefix for any hex values, otherwise I am treating any strings as decimal representations. without forcing the hex prefix, ie. `'15231'` can either be `15231` in decimal or `86577` in hex. so I figured I would just pass in the decimal representation as a string to make things simpler.

src/helpers/utility.js
dwalintukan: yes current average block time for testnet. I am planning on doing a follow up PR to grab the current average block time from the insight-api and using that instead of hardcoded numbers.

src/components/bodhi/createTopic.js
frankobe: @dwalintukan 2^32 is bit enough for unix timestamp. no need for string or hex for now

In the backend side it will be stored as int anyway so $lt etc can be performed

PR #118, opened 2018-02-05
Get average block time from Insight-API

dwalintukan:


src/network/httpRequest.js
```diff
@@ -0,0 +1,42 @@
...
+export function request(url, options) {
+  console.log('url:', url, 'options:', options);
+
+  return fetch(url, options)
+    .then(checkStatusCode)
+    .then(parseResponse);
+}
+
+function checkStatusCode(response) {
+  if (response.status >= 200 && response.status < 300) {
```
frankobe: 3xx can be redirect, not an error

src/network/httpRequest.js
```diff
@@ -0,0 +1,42 @@
...
+  return Promise.reject(new Error(response.statusText));
+}
+
+/**
+ * Parses the HTTP response returned by a network request.
+ * @param response {Object} A response from a network request.
+ * @return {Object} The parsed JSON from the request.
+ */
+function parseResponse(response) {
+  if (response.url === 'https://testnet.qtum.org/insight-api/statistics/total') {
```
frankobe: why do we need this? it returns a json

src/network/httpRequest.js
dwalintukan: Node-fetch seems to handle redirects. I'll research, but I think if it does redirect it should be returning the code of the redirect response.

src/network/httpRequest.js
dwalintukan: The response is it GZIP format so I need uncompress and read it with `response.text()` first, then parse with `JSON.parse()`.

Let me make this more generic and detect if it is a GZIP instead of using the url.

dwalintukan: Nevermind. This was user error (nothing to do with GZIP). I was trying to return `res.result` when that didn't exist for the insight-api. Fixed to return the correct payload.

PR #123, opened 2018-02-09
Call unregister sw in index

dwalintukan:
Call unregister for existing service workers to remove any caching that was previously registered in earlier versions.

### DIFF
--- a/src/registerServiceWorker.js
+++ b/src/registerServiceWorker.js
@@ -15,8 +15,6 @@ export default function register() {
       navigator.serviceWorker
         .register(swUrl)
         .then((registration) => {
-          console.debug('Registered service worker', registration);
-
           const reg = registration;
           reg.onupdatefound = () => {
             const installingWorker = registration.installing;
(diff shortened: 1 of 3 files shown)


src/registerServiceWorker.js
```diff
@@ -15,6 +15,8 @@ export default function register() {
       navigator.serviceWorker
         .register(swUrl)
         .then((registration) => {
+          console.debug('Registered service worker', registration);
```
frankobe: remove log?

src/registerServiceWorker.js
```diff
@@ -45,6 +47,7 @@ export function unregister() {
   if ('serviceWorker' in navigator) {
     navigator.serviceWorker.ready.then((registration) => {
       registration.unregister();
+      console.debug('Unregistered service worker');
```
frankobe: remove log?

src/registerServiceWorker.js
dwalintukan: guess not necessary. ok ill remove

src/registerServiceWorker.js
dwalintukan: removed

PR #141, opened 2018-02-21
Implement GraphQL mutation calls

dwalintukan:


src/network/graphDataStruct.js
```diff
@@ -74,11 +60,253 @@ const FIELD_MAPPINGS = {
...
+        version
+        senderAddress
+        oracleAddress
+        optionIdx
+        amount
+      }
+    }
+  `,
+
+  approve: `
```
frankhu-optimizely: `approve` wont be in a separate mutation call. It is a separate transaction

frankobe: `approve` wont be in a separate mutation call, but it will be a separate transaction

dwalintukan: gotcha, removed

dwalintukan:
@frankobe updated

PR #142, opened 2018-02-22
Fix padding on Create Event Results

liviavinci:


liviavinci:
screen shot 2018-02-22 at 8 00 24 am

dwalintukan:
Nice. I like the popup.

PR #143, opened 2018-02-22
Fix: Oracle page Radio button click animation is not circle

liviavinci:


liviavinci:
The bug is library level and is caused by excessive padding and height. The fix is a bit hacky but it's the only way.

src/style/oracle.less
```diff
@@ -0,0 +1,7 @@
+.cardVoting .body .ant-radio-group .ant-radio-wrapper .ant-radio {
+	padding: 0;
+	height: 14px;
+	margin-right: 24px;
+	position: relative;
+	top: 27px;
+}
```
dwalintukan: Add extra line at end of file (to get rid of this red thingy).

src/style/styles.less
```diff
@@ -16,4 +16,6 @@ ALL LESS FILES
 
 @import './cornerclock.less';
 
-@import './createTopic.less';
\ No newline at end of file
+@import './createTopic.less';
+
+@import './oracle.less';
```
dwalintukan: Add extra line at EOF

PR #144, opened 2018-02-22
add listener to network connection

liviavinci:


liviavinci:
screen shot 2018-02-22 at 1 03 56 pm
screen shot 2018-02-22 at 1 04 04 pm

PR #156, opened 2018-02-25
cleanup old layout usage; fill in theme and vars

liviavinci:


liviavinci:
FYI the variables.less is there only becuz the reorg is not entirely finished, it will be gone afterwards.
All styles will be in JSS and will be based on theme.js in the future.

I'll write a documentation on Frontify after prediction detail page makeover is done. I'm using prediction detail page as a demo to design and settle the structures and the rules.

PR #160, opened 2018-02-27
Localization

henryzbw:
image

Localization for the app, click the language icon to change language


src/appProvider.js
```diff
@@ -15,25 +15,44 @@ import { store, history } from './redux/store';
...
+              </ConnectedRouter>
+            </Provider>
+          </ApolloProvider>
+        </IntlProvider>
+      </MuiThemeProvider>
+    );
+  }
+}
+
+// AppProvider.propTypes = {
```
liviavinci: remove this?

PR #163, opened 2018-02-27
Use mutations for all transactions

dwalintukan:


liviavinci:
Looking good after moving redux back out.

PR #168, opened 2018-03-01
Bottom bar

liviavinci:


liviavinci:
screen shot 2018-03-01 at 5 28 56 pm
screen shot 2018-03-01 at 5 28 47 pm

src/languageProvider/entries/en-US.js
```diff
@@ -4,9 +4,10 @@ import enMessages from '../locales/en_US.json';
 
 const EnLang = {
   messages: {
-    'cornerclock.online': 'Online',
-    'cornerclock.offline': 'Offline',
-    'cornerclock.block': 'Current Block',
+    'bottombar.online': 'Online',
+    'bottombar.offline': 'Offline',
+    'bottombar.blockNum': 'Current Block Number',
+    'bottombar.blockTime': 'Current Block Time',
```
dwalintukan: use camelCase for the component name too: ie. `bottomBar.blockTime`

src/languageProvider/entries/zh-Hans-CN.js
```diff
@@ -4,9 +4,10 @@ import zhMessages from '../locales/zh-Hans.json';
 
 const ZhLan = {
   messages: {
-    'cornerclock.online': '在线',
-    'cornerclock.offline': '离线',
-    'cornerclock.block': '当前区块数',
+    'bottombar.online': '在线',
+    'bottombar.offline': '离线',
+    'bottombar.blockNum': '当前区块数',
+    'bottombar.blockTime': '当前区块时间',
```
dwalintukan: camelCase

src/components/BottomBar/index.js
```diff
@@ -29,41 +38,62 @@ class BottomBar extends React.PureComponent {
   }
 
   render() {
-    const { syncBlockNum, syncBlockTime } = this.props;
+    const { classes, syncBlockNum, syncBlockTime } = this.props;
+
+    return (
+      syncBlockTime && syncBlockTime ?
```
dwalintukan: syncBlockNum && syncBlockTime?

src/components/BottomBar/index.js
```diff
@@ -29,41 +38,62 @@ class BottomBar extends React.PureComponent {
...
+        <Paper className={classes.bottomBarWrapper}>
+          <Grid container>
+            {this.renderNetworkConnection()}
+            {this.renderBlockInfo()}
+          </Grid>
+        </Paper> : null
```
dwalintukan: do you think we should still show this bar even if the we cant get syncBlockTime/num? and maybe show error text: `Error getting block info`?

Also, the network connection text should be visible always

dwalintukan:
looks good!

src/components/BottomBar/index.js
liviavinci: this is to hide the bottom bar if data is not loaded

src/components/BottomBar/index.js
liviavinci: Let's hide the block number if it's not loaded. It's not nice to display error to users directly. And there is the loading phase so...

PR #169, opened 2018-03-01
Actionable items

dwalintukan:
Gets the `actionableItemCount` for the badge number in the Nav Bar My Activities badge:

This design does not have the number, but we can put it there.
image

Also refactored the Redux items and put the correct actions, sagas, reducers in the right spot.


src/redux/Topic/saga.js
```diff
@@ -35,17 +36,17 @@ export function* calculateWinningsRequestHandler() {
         type: actions.CALCULATE_WINNINGS_RETURN,
         value,
       });
-    } catch (error) {
+    } catch (err) {
       yield put({
         type: actions.CALCULATE_WINNINGS_RETURN,
-        value: { error: error.message ? error.message : '' },
+        error: err.message,
```
liviavinci: err.message or err?

src/redux/app/reducer.js
```diff
@@ -13,7 +13,7 @@ const initState = new Map({
   current: preKeys,
   walletAddrs: [],
   walletAddrsIndex: 0,
-  selected_wallet_address: 'wtf',
+  selectedWalletAddress: '',
```
liviavinci: lol good

src/redux/Topic/saga.js
dwalintukan: I am returning just the string currently. If we have any use later to have the entire `error` object, will refactor all the instances later.

I wanted to have a catch-all for all errors in the reducers.

src/redux/app/reducer.js
dwalintukan: 🤣

PR #186, opened 2018-03-06
change non-xl screen to full width

liviavinci:


liviavinci:
screen shot 2018-03-06 at 6 43 00 pm

PR #187, opened 2018-03-06
Wallet Transactions biz logic and some styling

dwalintukan:
**Added functionality for Deposit and Withdraw buttons.**
image

**Deposit Dialog**
image

**Withdraw Dialog**
image


src/components/TransactionSentDialog/index.js
```diff
@@ -0,0 +1,114 @@
...
+        body1: '',
+        body2: '',
+      };
+    }
+
+    return (
+      <Dialog
+        open={Boolean(txReturn)}
+        onClose={this.handleAlertClose}
+        aria-labelledby="alert-dialog-title"
```
liviavinci: delete?

src/components/TransactionSentDialog/index.js
```diff
@@ -0,0 +1,114 @@
...
+    this.getSuccessText = this.getSuccessText.bind(this);
+    this.getErrorText = this.getErrorText.bind(this);
+  }
+
+  render() {
+    const { classes, txReturn } = this.props;
+
+    let text;
+    if (txReturn) {
+      text = txReturn && txReturn.txid ? this.getSuccessText() : this.getErrorText();
```
liviavinci: can we have better names for this block of data?

src/components/TransactionSentDialog/index.js
```diff
@@ -0,0 +1,114 @@
...
+    const { intl, txReturn } = this.props;
+    return {
+      title: intl.formatMessage(messages.failureMsg),
+      body1: txReturn.error,
+      body2: '',
+    };
+  }
+
+  handleAlertClose() {
+    window.location.reload();
```
liviavinci: we can reload for now but will need a smoothing handling method in the future

src/components/TransactionSentDialog/styles.js
```diff
@@ -0,0 +1,7 @@
+const styles = (theme) => ({
+  bodyText1: {
```
liviavinci: better naming

src/languageProvider/locales/en_US.json
```diff
@@ -87,26 +87,35 @@
...
   "topbar.event": "Event",
   "topbar.events": "Events",
   "topbar.setting": "Setting",
   "topbar.voting": "Voting",
+  "transactionSentDialog.successMsg": "Success",
+  "transactionSentDialog.failureMsg": "Oops, something went wrong",
+  "transactionSentDialog.waitingMsg": "Transaction sent. Waiting for confirmations.",
   "walletHistory.transactionHistory": "Transaction History",
   "walletHistory.time": "Time",
-  "walletHistory.from": "From",
```
liviavinci: Can we keep a list of missing translation somewhere?

src/components/TransactionSentDialog/index.js
dwalintukan: done

src/components/TransactionSentDialog/index.js
dwalintukan: renamed to `textContent`

src/components/TransactionSentDialog/index.js
dwalintukan: ya i personally think we should not force reload, but then we should go refreshes on each page this is used. lets do this in a followup PR.

src/components/TransactionSentDialog/styles.js
dwalintukan: done

src/languageProvider/locales/en_US.json
dwalintukan: this wasn't deleted, just moved from `walletHistory.from` to `str.from`. we've been using `str.xxx` for generic strings that can be used in more than one place.

PR #194, opened 2018-03-07
detail page tx history done  Livia Zhang committed

liviavinci:


liviavinci:
screen shot 2018-03-07 at 6 26 37 pm

src/scenes/Oracle/components/PredictionTxHistory/index.js
```diff
@@ -0,0 +1,58 @@
...
+
+import { getLocalDateTimeString } from '../../../../helpers/utility';
+import styles from './styles';
+
+class PredictionTxHistory extends React.PureComponent {
+  render() {
+    const { classes, transactions, options } = this.props;
+
+    if (!transactions.length || !options.length) {
+      return null;
```
dwalintukan: do you think we should show the empty table regardless if they have txs or not? thinking it might be weird to have a whole section just pop up after they do a tx?

liviavinci: Mmm let's add a note instead of hiding the whole thing then.

dwalintukan: sounds good :dog:

PR #213, opened 2018-03-07
Wallet Transaction History biz logic

dwalintukan:
image


src/scenes/Wallet/components/Balances/index.jsx
```diff
@@ -254,10 +254,8 @@ class MyBalances extends React.PureComponent {
 
     return (
       <TableBody>
-        {data.map((item, index) => {
-          const className = index % 2 === 0 ? classes.tableRow : classNames(classes.tableRow, 'dark');
-
-          return (<TableRow key={item.address} className={className}>
+        {data.map((item, index) =>
+          (<TableRow key={item.address} selected={index % 2 !== 0} className={classes.tableRow}>
```
liviavinci: remove table row class?

src/scenes/Wallet/components/Balances/styles.js
```diff
@@ -22,24 +22,12 @@ const styles = (theme) => ({
...
-    border: theme.border,
-  },
-  tableHeader: {
-    height: theme.sizes.table.headerHeight,
-    background: theme.palette.background.grey,
-  },
   tableHeaderItemText: {
     fontWeight: theme.typography.fontWeightBold,
   },
   tableRow: {
```
liviavinci: i don't think we need this anymore since it's handled in the global theme overrides

src/scenes/Wallet/components/Balances/index.jsx
dwalintukan: removed

src/scenes/Wallet/components/Balances/styles.js
dwalintukan: removed and added padding to theme

PR #214, opened 2018-03-07
fix missed camelCase

henryzbw:
Change `bottomButton.` to `bottomButtonText. ` and re-run `npm run build:langs` to remove  unused ids


src/languageProvider/locales/en_US.json
```diff
@@ -93,36 +73,19 @@
...
-  "str.actions": "Actions",
-  "str.address": "Address",
+  "str.RAISE": "RAISED",
   "str.amount": "Amount",
-  "str.bot": "BOT",
-  "str.close": "Close",
-  "str.copy": "Copy",
-  "str.end": "Ended",
-  "str.fee": "Fee",
-  "str.from": "From",
```
dwalintukan: these shouldn't be deleted, they are being used by `MyWallet` pages. I think there is something wrong with your import script. It seems to be deleting ones that don't have translations.

PR #221, opened 2018-03-08
Disable action

dwalintukan:
- Refactor logic to enable/disable Oracle action button (Bet/Set/Vote/Finalize).
- Check for pending tx on that Oracle
- Added message to show why the button is disabled #243 

image

image

image


src/components/EventCardsGridContainer/index.js
```diff
@@ -35,7 +35,7 @@ const messages = defineMessages({
     defaultMessage: 'Finalize Result',
   },
   withdraw: {
-    id: 'bottomButtonText.withdraw',
+    id: 'str.withdraw',
```
frankobe: does `str.withdraw` means `withdraw` literally? if so we can use `withdraw` directly. @henryzbw what do u think

src/scenes/Oracle/index.js
```diff
@@ -235,101 +250,58 @@ class OraclePage extends React.Component {
     this.props.getOracles([
       { topicAddress: this.state.topicAddress },
     ], undefined);
-    // TODO (LIVIA): NEED TO TXS FOR THE ENTIRE TOPIC
+
     this.props.getTransactions([
-      { oracleAddress: this.state.address },
+      { topicAddress: this.state.topicAddress },
```
frankobe: is this dependable on every transaction must hv the topicAddress attribute?

src/components/EventCardsGridContainer/index.js
dwalintukan: I'm ok with `withdraw` directly instead of `str.withdraw`. But in terms of key organization, if we have the `str.` in front of it, all of the global strings will be together.

src/scenes/Oracle/index.js
dwalintukan: yup. only the Oracle and Topic detail pages will we query by `topicAddress`.

src/scenes/Oracle/index.js
```diff
@@ -157,6 +140,39 @@ class OraclePage extends React.Component {
...
+    // Did not select an option
+    if (currentOptionIdx === -1) {
+      return true;
+    }
+
+    // Did not enter an amount
+    if ((voteAmount === 0 || Number.isNaN(voteAmount)) && config.predictionAction.showAmountInput) {
+      return true;
+    }
+
```
liviavinci: should we also check if the amount input is smaller than my balance?

src/scenes/Event/scenes/oracle.js
```diff
@@ -182,6 +165,39 @@ class OraclePage extends React.Component {
...
+      return true;
+    }
+
+    // Already have a pending tx for this Oracle
+    const pendingTxs = _.filter(transactions, { oracleAddress: address, status: TransactionStatus.Pending });
+    if (pendingTxs > 0) {
+      return true;
+    }
+
+    // Did not select an option
```
liviavinci: This will cause a but in finalize stage

src/scenes/Event/scenes/oracle.js
```diff
@@ -182,6 +165,39 @@ class OraclePage extends React.Component {
...
+    if (currentOptionIdx === -1) {
+      return true;
+    }
+
+    // Did not enter an amount
+    if ((voteAmount === 0 || Number.isNaN(voteAmount)) && config.predictionAction.showAmountInput) {
```
liviavinci: This will cause a but in finalize and set result stage

src/scenes/Event/scenes/oracle.js
```diff
@@ -182,6 +165,39 @@ class OraclePage extends React.Component {
...
+      return true;
+    }
+
```
liviavinci: also we will want to handle the set/finalize stage. If a result is already set. We should allow another set. If the event is already finalized, we shouldn't allow another finalize, and should go to topic page.

frankobe: @liviavinci  if a result is already set, we shouldn't allow another set

dwalintukan:
will make changes when #256 gets merged in first. I want to include the check for amount < balance.

src/scenes/Oracle/index.js
dwalintukan: will take care of this when #256 is merged in. don't want to fix it with old logic.

src/scenes/Event/scenes/oracle.js
dwalintukan: fixed

src/scenes/Event/scenes/oracle.js
dwalintukan: this `if` will return false during those two stages because of `config.predictionAction.showAmountInput`

src/scenes/Event/scenes/oracle.js
dwalintukan: ditto. can only setResult once. regarding the finalize route to topic page, we have another issue open for this. will fix in that other PR.

PR #250, opened 2018-03-09
change navbar icon and removed some unused icon imports

liviavinci:


liviavinci:
screen shot 2018-03-09 at 6 11 25 pm

PR #254, opened 2018-03-09
made some small refactors, helping to get familiar with codebase

alex-cory:
- scoping of global variables moved to class
- shortened code
- added new syntax (fat arrow functions, class properties, etc.)

### DIFF
--- a/src/appProvider.js
+++ b/src/appProvider.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Component } from 'react';
 import { ApolloProvider } from 'react-apollo';
 import { ConnectedRouter } from 'react-router-redux';
 import { IntlProvider } from 'react-intl';
@@ -15,21 +15,17 @@ import { store, history } from './redux/store';
 
 import '../src/style/styles.less';
 
-const locales = [AppLocale.en, AppLocale.zh];
-class AppProvider extends React.Component {
+export default class AppProvider extends Component {
+  locales = [AppLocale.en, AppLocale.zh]
   state = {
     locale: localStorage.getItem('localindex') || 0,
-  }
-
-  componentDidMount() {
-    moment.locale(locales[this.state.locale].momentlocale);
   }
 
   langHandler = () => {
     this.setState({
       locale: (this.state.locale + 1) % 2,
     }, () => {
-      moment.locale(locales[this.state.locale].momentlocale);
+      moment.locale(this.locales[this.state.locale].momentlocale);
       localStorage.setItem('localindex', this.state.locale);
     });
   }
@@ -37,8 +33,8 @@ class AppProvider extends React.Component {
   render() {
     return (
       <MuiThemeProvider theme={bodhiTheme}>
-        <LocaleProvider locale={locales[this.state.locale].antd}>
-          <IntlProvider locale={locales[this.state.locale].locale} messages={locales[this.state.locale].messages}>
+        <LocaleProvider locale={this.locales[this.state.locale].antd}>
+          <IntlProvider locale={this.locales[this.state.locale].locale} messages={this.locales[this.state.locale].messages}>
             <ApolloProvider client={graphClient}>
               <Provider store={store}>
                 <ConnectedRouter history={history}>
@@ -55,5 +51,3 @@ class AppProvider extends React.Component {
     );
   }
 }
-
-export default AppProvider;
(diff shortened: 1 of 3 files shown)


src/appProvider.js
```diff
@@ -15,16 +15,13 @@ import { store, history } from './redux/store';
...
-const locales = [AppLocale.en, AppLocale.zh];
-class AppProvider extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = { locale: defaultidx };
-    this.langHandler = this.langHandler.bind(this);
+export default class AppProvider extends Component {
+  locales = [AppLocale.en, AppLocale.zh]
+  state = {
+    locale: localStorage.getItem('localindex') || 0,
```
dwalintukan: is this a built-in state var? `localindex`?

src/appProvider.js
```diff
@@ -15,16 +15,13 @@ import { store, history } from './redux/store';
...
   }
 
-  langHandler() {
+  langHandler = () => {
```
dwalintukan: is this necessary for this to work? i'd prefer to keep it as functions itself versus a variable that holds a function.

unless there is some specific benefit for doing it this way?

src/scenes/App/globalHub.jsx
```diff
@@ -11,16 +11,30 @@ import AppConfig from '../../config/app';
...
 let syncInfoInterval;
 
-class GlobalHub extends React.PureComponent {
-  constructor(props) {
-    super(props);
-
-    this.state = {
-    };
+class GlobalHub extends PureComponent {
+  static propTypes = {
```
dwalintukan: cool this is much cleaner 👍

src/scenes/App/globalHub.jsx
```diff
@@ -67,7 +81,7 @@ class GlobalHub extends React.PureComponent {
     return null;
   }
 
-  subscribeSyncInfo() {
+  subscribeSyncInfo = () => {
```
dwalintukan: same comment, is it necessary?

src/scenes/App/globalHub.jsx
```diff
@@ -82,19 +96,15 @@ class GlobalHub extends React.PureComponent {
     });
   }
 
-  pollSyncInfo() {
+  pollSyncInfo = () => {
```
dwalintukan: ^

src/scenes/App/globalHub.jsx
```diff
@@ -82,19 +96,15 @@ class GlobalHub extends React.PureComponent {
...
     syncInfoInterval = setInterval(getSyncInfo, AppConfig.intervals.syncInfo);
   }
 
-  updateBalances(walletAddresses) {
-    const {
-      initSyncing,
-      listUnspent,
-      getBotBalance,
-    } = this.props;
+  updateBalances = (walletAddresses) => {
```
dwalintukan: ^

src/appProvider.js
alex-cory: I'm not exactly sure why we're using this. My guess would be maybe this is used for electron.

src/appProvider.js
alex-cory: It autobinds this to the class so you don't have to do it in the constructor or wherever the method is called.  It's just a fat arrow function definition, not a variable.
Old:
```javascript
class AppProvider extends Component {
  constructor(props) {
    this.state = { ... }
    this.langHandler = this.langHandler.bind(this)
  }
  langHandler() {
  }
}
```
New:
```javascript
class AppProvider extends Component {
  state = {}
  langHandler = () => {
  }
}
```

alex-cory: You can see how repetitive that gets if you have like 5 methods and you have to bind all of them in the constructor or have callbacks in your render method like `() => this.langHandler`

src/appProvider.js
dwalintukan: this is actually for our localization

src/appProvider.js
dwalintukan: gotcha, this makes it much easier than to bind everything 👍

dwalintukan:
Looks good after a rebase. please make sure to `rebase` by `git pull --rebase origin master`. And **not** `merge` when resolving conflicts.

src/appProvider.js
henryzbw: it's used to set the default language for the browser users before. Let them change language one time, and in the future, when they open our app, it will automatically show the language they choose last time. But it seems not work in electron?

dwalintukan: i think its fine for electron. havent seen any issues yet.

PR #258, opened 2018-03-10
Show mutated/unconfirmed Oracle in Bet dashboard

dwalintukan:
image

image


liviavinci:
Please add a pending info label

dwalintukan:
@liviavinci do you mean on the EventCard?

liviavinci:
It would be nice if we have it both on the card and the detail page

dwalintukan:
@liviavinci Added chip and changed text in detail page.

image

src/components/EventCard/index.js
```diff
@@ -14,6 +15,22 @@ import styles from './styles';
 import { getLocalDateTimeString, getEndTimeCountDownString } from '../../helpers/utility';
 
 class EventCard extends React.PureComponent {
+  static propTypes = {
```
liviavinci: why static?

src/languageProvider/locales/en_US.json
```diff
@@ -57,6 +57,7 @@
   "oracle.betStartTimeDisabledText": "The betting start time has not started yet.",
   "oracle.consensus": "Consensus Threshold {value}. This value indicates the amount of BOT needed to set the result.",
   "oracle.enterAmountDisabledText": "You have not entered a valid amount.",
+  "oracle.eventUnconfirmed": "This created Event is unconfirmed. You cannot interact with it until it is confirmed by the blockchain.",
```
liviavinci: Please add corresponding vars in chinese json

src/components/EventCard/index.js
dwalintukan: propTypes are static items as they are all declared the same way across all instances. the previous method of declaring them is the same way:
EventCard.propTypes is declaring them in a static way.

src/languageProvider/locales/en_US.json
dwalintukan: ran the script

PR #260, opened 2018-03-11
Change font size of event info

dwalintukan:


src/scenes/Event/components/EventInfo/styles.js
```diff
@@ -10,7 +10,8 @@ const styles = (theme) => ({
     textOverflow: 'ellipsis',
     whiteSpace: 'nowrap',
     overflow: 'hidden',
-    paddingLeft: theme.padding.lg.px,
+    paddingLeft: theme.padding.md.px,
+    fontSize: theme.typography.fontSize,
```
liviavinci: why is this necessary? theme.typography.fontSize is just the default font size right?

dwalintukan: the Result Setter text is being ellipsized. i changed the font size because i wanted it black color still, just smaller text size. suggestions?

liviavinci: oh i did that ellipsis thing on purpose but now as i think of it, it doesn't make much sense. Can you remove textOverflow: 'ellipsis', whiteSpace: 'nowrap',overflow: 'hidden',?

src/scenes/Event/components/EventInfo/styles.js
```diff
@@ -10,7 +10,8 @@ const styles = (theme) => ({
     textOverflow: 'ellipsis',
```
liviavinci: remove

dwalintukan: wont this just cutoff the end of the text and some of the letters will still be missing?

liviavinci: well if we remove textOverflow: 'ellipsis', whiteSpace: 'nowrap',overflow: 'hidden', and add "word-wrap:break-word;". If the text is more than the div width, it will just be two lines

dwalintukan: i'm not sure if we should show a qtum address on two lines. think that feels kind of confusing for the user. can we adjust the grid of the main content to give the side content more room?

liviavinci: but the thing is, no matter how you adjust that, if the user drags the window into a smaller size, you'll have the same problem.

dwalintukan: true, change made.

src/scenes/Event/components/EventInfo/styles.js
dwalintukan: done. are you ok with the text size change or you think it should be big?

src/scenes/Event/components/EventInfo/styles.js
```diff
@@ -7,10 +7,9 @@ const styles = (theme) => ({
   },
   eventInfo: {
     marginTop: theme.padding.unit.px,
-    textOverflow: 'ellipsis',
-    whiteSpace: 'nowrap',
-    overflow: 'hidden',
-    paddingLeft: theme.padding.lg.px,
+    'word-wrap': 'break-word',
```
liviavinci: wordWrap: 'break-word',

dwalintukan: fixed

src/scenes/Event/components/EventInfo/styles.js
liviavinci: I'm ok with the smaller font size since it would be nice to make the address in one line on most of the laptops

PR #261, opened 2018-03-12
add link when no badge is shown

liviavinci:


liviavinci:
that was a quite stupid bug

dwalintukan:
😹

PR #262, opened 2018-03-12
display the correct current tab in navbar

liviavinci:


liviavinci:
screen shot 2018-03-12 at 4 20 28 pm

src/components/NavBar/index.js
```diff
@@ -39,13 +41,37 @@ class NavBar extends React.PureComponent {
...
+            <Button
+              data-index={EventStatus.Bet}
+              className={classNames(
+                classes.navEventsButton,
+                match.path === '/' ? 'selected' : '',
+              )}
+            >
+              QTUM Prediction
+            </Button>
+            { this.renderCurrentTabArrow('/') }
```
dwalintukan: what do you think about building a button component that has the built in arrow?
https://material-ui-next.com/demos/buttons/#complex-buttons

src/scenes/App/router.js
```diff
@@ -16,37 +18,37 @@ class AppRouter extends React.Component {
         <Route
           exact
           path={`${url}/`}
-          component={asyncComponent(() => import('../Dashboard/index'))}
+          component={asyncComponent(() => import('../Dashboard/index'), langHandler)}
```
dwalintukan: is it possible to inject this directly into the component vs having to pass it?

PR #264, opened 2018-03-12
Add address picker to Create Event page

dwalintukan:
image

image


liviavinci:
Please add a screenshot for UI changes

dwalintukan:
oops sorry, done

liviavinci:
do you think we are better of with an downward arrow instead of a plus icon on the right of the input?

dwalintukan:
@liviavinci doesn't a down arrow assume that it will be a drop down?

src/components/SelectAddressDialog/index.jsx
```diff
@@ -0,0 +1,58 @@
...
+      dialogVisible,
+      walletAddresses,
+    } = this.props;
+
+    return (
+      <Dialog
+        open={dialogVisible}
+        onClose={this.handleClose}
+      >
+        <DialogTitle>Select Centralized Oracle</DialogTitle>
```
liviavinci: let's call it result setter and please handle the localization

liviavinci:
Good to go after the above comment and changing the icon

src/components/SelectAddressDialog/index.jsx
dwalintukan: done

PR #265, opened 2018-03-12
Activities count

liviavinci:


dwalintukan:
@liviavinci I see you `merged` instead of rebasing. To keep the commit history clean we need to rebase. I can walk you through the steps if you are unsure how.

src/redux/Graphql/saga.js
```diff
@@ -118,29 +118,55 @@ function processTransaction(tx) {
...
+      totalCount += result.length;
+
+      const oracleFinalizeFilters = [
+        { token: Token.Bot, status: OracleStatus.WaitResult },
+      ];
+      result = yield call(queryAllOracles, oracleFinalizeFilters);
+      countByStatus[EventStatus.Finalize] = result.length;
+      totalCount += result.length;
+
+      console.log({
```
dwalintukan: remove

src/redux/Graphql/saga.js
```diff
@@ -118,29 +118,55 @@ function processTransaction(tx) {
...
+        },
       });
     } catch (err) {
       console.log(err);
       yield put({
         type: actions.GET_ACTIONABLE_ITEM_COUNT_RETURN,
-        value: 0,
+        value: {
+          totalCount: 0,
+          countByStatus,
```
dwalintukan: why `totalCount: 0`? It possible the error can be caught after a request or two came back already (and incremented the totalCount).

But that being said, I think we need to have some sort of default error catcher in the reducer.

PR #278, opened 2018-03-13
display balance in address selector

liviavinci:


liviavinci:
screen shot 2018-03-13 at 5 06 07 pm

PR #279, opened 2018-03-13
Use syncInfo balances instead of polling

dwalintukan:


src/redux/App/reducer.js
```diff
@@ -13,53 +14,58 @@ const initState = new Map({
...
+        });
+      });
 
-      // Update the bot balance of the existing addresses or add in existing addresses but with 0 qtum balance.
-      // This is for good UX so the user won't wonder why their old address which they did a sendtoaddress
-      // is no longer on the list of addresses.
-      const existingAddresses = state.get('walletAddresses');
-      _.each(existingAddresses, (addressObj) => {
+      // Add old addresses to new list if not found
+      const oldAddresses = state.get('walletAddresses');
```
frankobe: Is it because after usage, the address balance becomes 0 and it wont be included in listunspent?

dwalintukan: correct. this should be solved later with the `accounts` db when we add that to keep track of all used addresses.

PR #280, opened 2018-03-13
Withdraw detail balances and more

dwalintukan:
- show correct withdraw balances and contributions per address
- fix address switcher
- add betBalances and voteBalances redux logic
- localization

image


frankobe:
@dwalintukan looks awesome

PR #285, opened 2018-03-14
fix the bug when sort is messed up when switching between dashboard tabs

liviavinci:


src/scenes/Dashboard/components/TopActions/index.js
```diff
@@ -69,14 +61,18 @@ class TopActions extends Component {
 
 TopActions.propTypes = {
```
dwalintukan: please declare as `static` and move inside class structure

src/scenes/Dashboard/components/TopActions/index.js
```diff
@@ -69,14 +61,18 @@ class TopActions extends Component {
...
   classes: PropTypes.object.isRequired,
+  sortBy: PropTypes.string,
   sortOrderChanged: PropTypes.func,
 };
 
 TopActions.defaultProps = {
```
dwalintukan: static and move to class

src/scenes/Dashboard/index.js
```diff
@@ -33,7 +33,7 @@ Dashboard.propTypes = {
 };
 
 Dashboard.defaultProps = {
-  sortBy: undefined,
+  sortBy: SortBy.Ascending,
```
dwalintukan: do you think we should have this default in the `reducer` so we dont have to declare a default for each class that uses it?

liviavinci: need to think about this. cuz we not only need the sort order for request but also for display. let's do a reorg later?

PR #330, opened 2018-03-16
Transfer and activity history sorting fixes

dwalintukan:
image

### DIFF
--- a/src/scenes/Activities/scenes/EventHistory/index.js
+++ b/src/scenes/Activities/scenes/EventHistory/index.js
@@ -36,22 +36,11 @@ class EventHistory extends React.Component {
     syncBlockTime: undefined,
   };
 
-  constructor(props) {
-    super(props);
-
-    this.state = {
-      transactions: [],
-      order: 'desc',
-      orderBy: 'createdTime',
-    };
-
-    this.getTableHeader = this.getTableHeader.bind(this);
-    this.getTableRows = this.getTableRows.bind(this);
-    this.createSortHandler = this.createSortHandler.bind(this);
-    this.handleSorting = this.handleSorting.bind(this);
-    this.executeTxsRequest = this.executeTxsRequest.bind(this);
-    this.onEventLinkClicked = this.onEventLinkClicked.bind(this);
-  }
+  state = {
+    transactions: [],
+    order: 'desc',
+    orderBy: 'createdTime',
+  };
 
   componentWillMount() {
     this.executeTxsRequest();
@@ -110,7 +99,7 @@ class EventHistory extends React.Component {
     );
   }
 
-  executeTxsRequest() {
+  executeTxsRequest = () => {
     this.props.getTransactions(
       [
         { type: TransactionType.CreateEvent },
@@ -124,9 +113,9 @@ class EventHistory extends React.Component {
       ],
       undefined
     );
-  }
+  };
 
-  getTableHeader() {
+  getTableHeader = () => {
     const headerCols = [
       {
         id: 'createdTime',
@@ -193,7 +182,7 @@ class EventHistory extends React.Component {
         </TableRow>
       </TableHead>
     );
-  }
+  };
 
   getSortableCell = (column) => {
     const { order, orderBy } = this.state;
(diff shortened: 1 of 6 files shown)


src/scenes/Wallet/components/History/index.jsx
```diff
@@ -17,18 +18,27 @@ import { getShortLocalDateTimeString, decimalToSatoshi } from '../../../../helpe
...
+    txReturn: PropTypes.object,
+    syncBlockNum: PropTypes.number.isRequired,
+  };
+
+  static defaultProps = {
+    getTransactionsReturn: [],
+    txReturn: undefined,
+  };
+
   constructor(props) {
```
alex-cory: we don't need the constructor anymore. We can just do:
```javascript
state = {
  data: [],
  ...
}
```

src/scenes/Activities/scenes/EventHistory/index.js
```diff
@@ -40,9 +40,9 @@ class EventHistory extends React.Component {
     super(props);
```
alex-cory: We don't need the constructor here either and we can make these methods fat arrow functions and remove the bindings in the constructor.

src/scenes/Activities/scenes/EventHistory/index.js
```diff
@@ -117,90 +127,110 @@ class EventHistory extends React.Component {
...
-            </TableCell>
-          ))}
+          {headerCols.map((column) => column.sortable ? this.getSortableCell(column) : this.getNonSortableCell(column))}
         </TableRow>
       </TableHead>
     );
   }
 
-  getTableRows(transactions) {
+  getSortableCell = (column) => {
```
alex-cory: Just from experience, this style of splitting the dom into various functions is being moved away from. Instead, putting them in separate components keeps things more concise since dom elements *are* components, not methods. i.e.
```javascript
class EventHistory extends Component {
  ...
  render() {
    return (
      <TableHead>
        <TableRow>
          {headerCols.map((column) => column.sortable ? (
            <SortableTableCell {...column} {...this.state} onClick={this.createSortHandler(column.id)} />
           ) : (
             <NonSortableTableCell {...column} {...this.state} />
           ))}
        </TableRow>
      </TableHead>
    );
  }
}

const SortableTableCell = ({ id, numeric, name, nameDefault, order, orderBy, onClick }) => (
  <TableCell key={id} numeric={numeric} sortDirection={orderBy === id ? order : false}>
    <Tooltip
      title={<FormattedMessage id="str.sort" defaultMessage="Sort" />}
      enterDelay={Config.intervals.tooltipDelay}
      placement={numeric ? 'bottom-end' : 'bottom-start'}
    >
      <TableSortLabel active={orderBy === id} direction={order} onClick={onClick}>
        <FormattedMessage id={name} default={nameDefault} />
      </TableSortLabel>
    </Tooltip>
  </TableCell>
)

const NonSortableTableCell = ({ id, numeric, name, nameDefault, onClick }) => (
  <TableCell key={id} numeric={numeric}>
    <FormattedMessage id={name} default={nameDefault} />
  </TableCell>
)
```

src/scenes/Activities/scenes/EventHistory/index.js
dwalintukan: fixed. changed all to fat arrows and moved state outside

src/scenes/Activities/scenes/EventHistory/index.js
dwalintukan: i tried doing this but it threw an error with having to define the props for id, numeric, etc..

src/scenes/Wallet/components/History/index.jsx
```diff
@@ -144,14 +168,14 @@ class WalletHistory extends React.Component {
         </TableRow>
       </TableHead>
     );
-  }
+  };
 
   createSortHandler = (property) => (event) => {
     this.handleSorting(event, property);
```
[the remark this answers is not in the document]
dwalintukan: This doesn't seem to work. There needs to be this intermediary func for some reason.

PR #333, opened 2018-03-18
create event redesign

liviavinci:


liviavinci:
not ready for review

liviavinci:
screen shot 2018-03-19 at 2 22 12 am

liviavinci:
screen shot 2018-03-19 at 2 49 03 am

liviavinci:
screen shot 2018-03-19 at 2 50 45 am

henryzbw:
Please change the time format to 24-hr.
And add the localization for result setter placeholder.

liviavinci:
The CreateEvent file is quite long, will do a reorg/breakdown in my next PR.

liviavinci:
@henryzbw unfortunately the 24-hr format is not feasible since it's a component from material ui and 24hr/12hr mode is not customizable. will fix the localization thing.

src/scenes/Dashboard/components/TopActions/index.js
```diff
@@ -14,10 +14,15 @@ import Card from 'material-ui/Card';
...
 
 import dashboardActions from '../../../../redux/Dashboard/actions';
+import CreateEvent from '../../../../scenes/CreateEvent/index';
 import { SortBy } from '../../../../constants';
 import styles from './styles';
 
 class TopActions extends Component {
+  state = {
+    createDialogOpen: false,
+  };
```
alex-cory: does this work? Pretty sure this will throw an error with the `;` since  this is a class property.

dwalintukan: actually I believe it throws a lint error if you don't have the `;`. Even fat arrow funcs need the `;`.

src/scenes/CreateEvent/index.js
```diff
@@ -0,0 +1,652 @@
...
+            <Grid container>
+              <Grid item xs={3}>
+                {intl.formatMessage(messages.resultSetter)}
+              </Grid>
+              <Grid item xs={9}>
+                <Field
+                  required
+                  fullWidth
+                  name="resultSetter"
+                  placeholder="Enter the address of result setter or select your own address"
```
dwalintukan: Use localized string. And, grammar change: `Enter the address of the result setter or select your own address`

src/scenes/CreateEvent/index.js
```diff
@@ -0,0 +1,652 @@
...
+              walletAddresses={walletAddresses}
+              onClosed={this.onSelectResultSetterAddressDialogClosed}
+            />
+          </DialogContent>
+          <DialogActions>
+            <Button color="primary" onClick={onClose}>
+              {intl.formatMessage(messages.cancel)}
+            </Button>
+            <Button type="submit" color="primary" disabled={submitting} variant="raised">
+              {intl.formatMessage(messages.publish)}
```
dwalintukan: fyi, I think you can use `<FormattedMessage>` here

src/scenes/CreateEvent/index.js
```diff
@@ -0,0 +1,652 @@
...
+    setLastUsedAddress: (address) => dispatch(appActions.setLastUsedAddress(address)),
+    changeFormFieldValue: (field, value) => dispatch(change(FORM_NAME, field, value)),
+  };
+}
+
+const fields = [
+  'name',
+  'outcomes',
+  'resultSetter',
+  'creatorAddress',
```
dwalintukan: if these fields are reused, make class constants for them.

src/scenes/CreateEvent/index.js
```diff
@@ -0,0 +1,652 @@
...
+    }
+  });
+
+  // check dates
+  const bettingStartTime = moment(values.bettingStartTime);
+  const bettingEndTime = moment(values.bettingEndTime);
+  const resultSettingStartTime = moment(values.resultSettingStartTime);
+  const resultSettingEndTime = moment(values.resultSettingEndTime);
+
+  if (bettingStartTime.unix() > bettingEndTime.unix()) {
```
dwalintukan: also, we should make sure they technically give at least a 1 block difference between start and end times. doesn't make sense for them to have a start to end time to be only like x seconds as that may fall in the same block.

suggest we make it a min. gap between start and end times by 30 mins. That way we know there will be blocks in between. even thought I think 30 mins is short too, but we should let the user choose.

src/scenes/CreateEvent/index.js
```diff
@@ -0,0 +1,652 @@
...
+    errors.bettingEndTime = intl.formatMessage(messages.validBetEnd);
+  }
+
+  if (bettingEndTime.unix() > resultSettingStartTime.unix()) {
+    errors.resultSettingStartTime = intl.formatMessage(messages.validResultSetStart);
+  }
+
+  if (resultSettingStartTime.unix() > resultSettingEndTime.unix()) {
```
dwalintukan: same as above. suggest we make a min gap of like 30 mins between start and end times.

src/scenes/Dashboard/components/TopActions/index.js
liviavinci: ditto deric

src/languageProvider/locales/en_US.json
```diff
@@ -118,7 +125,7 @@
   "str.notEnoughBot": "You don't have enough BOT",
   "str.notEnoughQtum": "You don't have enough QTUM",
   "str.ok": "OK",
-  "str.outcome": "OUTCOMES",
+  "str.outcomes": "Outcomes",
```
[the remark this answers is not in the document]
liviavinci: I think we unified all results to outcomes in the redesign.

src/scenes/CreateEvent/index.js
```diff
@@ -0,0 +1,652 @@
...
+    selectAddressDialogVisibility: false,
+  };
+
+  validateTitleLength = (value) => {
+    const { intl } = this.props;
+    let hexString = _.isUndefined(value) ? '' : value;
+
+    // Remove hex prefix for length validation
+    hexString = Web3Utils.toHex(hexString).slice(2);
+    if (hexString && hexString.length > MAX_LEN_EVENTNAME_HEX) {
```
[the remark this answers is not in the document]
liviavinci: this is handled in the bottom general required check

src/scenes/CreateEvent/index.js
```diff
@@ -0,0 +1,652 @@
...
+      return intl.formatMessage(messages.datePast);
+    }
+
+    return null;
+  };
+
+  validateResultLength = (value) => {
+    const { intl } = this.props;
+
+    if (!value) {
```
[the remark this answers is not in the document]
liviavinci: actually by using the redux form, undefined / null situation doesn't exist

src/scenes/CreateEvent/index.js
```diff
@@ -0,0 +1,652 @@
...
+    meta: { touched, error },
+    ...custom
+  }) => {
+    // calculate block num if input value is not empty
+    let blockNum = '';
+
+    if (!input.value || input.value !== '') {
+      const {
+        syncBlockNum,
+        averageBlockTime,
```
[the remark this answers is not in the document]
liviavinci: I'm actually using

import { defaults } from '../../config/app';

  static defaultProps = {
    averageBlockTime: defaults.averageBlockTime,
  };

will this work?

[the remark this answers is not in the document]
dwalintukan: ya this should work. when the return from Insight comes back it will override the props.

src/scenes/Dashboard/components/TopActions/index.js
alex-cory: weird... I'm surprised since you can't actually put logic inside a class, only methods inside a class. I don't have any lint errors for instance with:
```
state = {
  ...,
}
```

src/scenes/Dashboard/components/TopActions/index.js
```diff
@@ -33,24 +38,32 @@ class TopActions extends Component {
     super(props);
 
     this.onSortOptionSelected = this.onSortOptionSelected.bind(this);
+    this.onCreateDialogClose = this.onCreateDialogClose.bind(this);
+    this.onCreateDialogOpen = this.onCreateDialogOpen.bind(this);
```
alex-cory: also, we should make these fat arrow functions, although I know you said you'd do that later :)

PR #342, opened 2018-03-19
Create event bot staking

dwalintukan:
Merge when this one is merged: https://github.com/bodhiproject/bodhi-graphql/pull/154

### DIFF
--- a/src/redux/Topic/saga.js
+++ b/src/redux/Topic/saga.js
@@ -6,6 +6,39 @@ import { request } from '../../network/httpRequest';
 import { satoshiToDecimal } from '../../helpers/utility';
 import Routes from '../../network/routes';
 
+export function* getEventEscrowAmountHandler() {
+  yield takeEvery(actions.GET_EVENT_ESCROW_AMOUNT, function* getEventEscrowAmountRequest(action) {
+    try {
+      const {
+        senderAddress,
+      } = action.params;
+
+      const options = {
+        method: 'POST',
+        body: JSON.stringify({
+          senderAddress,
+        }),
+        headers: { 'Content-Type': 'application/json' },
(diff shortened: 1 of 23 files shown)


src/redux/Topic/saga.js
```diff
@@ -6,6 +6,40 @@ import { request } from '../../network/httpRequest';
...
+      const options = {
+        method: 'POST',
+        body: JSON.stringify({
+          senderAddress,
+        }),
+        headers: { 'Content-Type': 'application/json' },
+      };
+
+      const result = yield call(request, Routes.eventEscrowAmount, options);
+      const eventEscrowAmount = satoshiToDecimal(result[0]);
```
liviavinci: why [0]? in order words, why is result an array?

src/scenes/CreateEvent/index.js
```diff
@@ -120,23 +120,26 @@ class CreateEvent extends React.Component {
   static propTypes = {
     classes: PropTypes.object.isRequired,
     fields: PropTypes.array.isRequired,
-    open: PropTypes.bool.isRequired,
-    onClose: PropTypes.func.isRequired,
+    walletAddresses: PropTypes.array.isRequired,
+    setLastUsedAddress: PropTypes.func.isRequired,
```
liviavinci: this is not used right?

src/redux/Topic/saga.js
dwalintukan: The key is actually `"0"` which corresponds to the first return param.

src/scenes/CreateEvent/index.js
dwalintukan: yup you are right, removed

PR #361, opened 2018-03-20
fixed logic for stepper

alex-cory:


src/helpers/cardInfoUtil.js
```diff
@@ -52,12 +52,12 @@ class CardInfoUtil {
       },
       {
         title: BETTING,
-        description: `${getLocalDateTimeString(cOracle.startTime)} 
+        description: `${getLocalDateTimeString(cOracle.startTime)}
```
dwalintukan: does this next line work? I've had trouble where putting it on the next line didn't show the space that is needed.

src/scenes/Event/scenes/oracle.js
```diff
@@ -357,7 +356,7 @@ class OraclePage extends React.Component {
             message: intl.formatMessage(messages.eventUnconfirmed),
           },
         };
-      } else if (token === Token.Qtum && status === OracleStatus.Voting) {
```
dwalintukan: revert this change. we need `Token` to differentiate between a Centralized Oracle vs Decentralized Oracle. 

`status === voting` can be both for COracle and DOracle. (`Voting` is a synonym for `Betting`. Frank just decided to reuse the status.)

src/scenes/Event/scenes/oracle.js
```diff
@@ -371,7 +370,7 @@ class OraclePage extends React.Component {
             btnText: <FormattedMessage id="str.bet" defaultMessage="Bet" />,
           },
         };
-      } else if (token === Token.Qtum && (status === OracleStatus.WaitResult || status === OracleStatus.OpenResultSet)) {
```
dwalintukan: revert, same comment as above. the status is reused for both COracle and DOracle

src/scenes/Event/scenes/oracle.js
```diff
@@ -389,7 +388,7 @@ class OraclePage extends React.Component {
             message: intl.formatMessage(messages.setResultExplanation),
           },
         };
-      } else if (token === Token.Bot && status === OracleStatus.Voting) {
```
dwalintukan: revert

src/scenes/Event/scenes/oracle.js
```diff
@@ -407,7 +406,7 @@ class OraclePage extends React.Component {
             message: intl.formatMessage(messages.voteExplanation),
           },
         };
-      } else if (token === Token.Bot && status === OracleStatus.WaitResult) {
```
dwalintukan: revert

src/scenes/Event/scenes/oracle.js
dwalintukan: it looks like this entire block is wrapped in `if (oracle && oracle.token === Token.Qtum)` which is incorrect. 

`else if (token === Token.Bot && status === OracleStatus.Voting)` and `else if (token === Token.Bot && status === OracleStatus.WaitResult)` will never be reached (the last two cases).

src/helpers/cardInfoUtil.js
alex-cory: I didn't actually change this. I think it was my editor just removing spaces at the end of lines.

src/scenes/Event/scenes/oracle.js
alex-cory: sorry about that. My bad 😵

PR #369, opened 2018-03-22
Dont show result for non-validated voting rounds

dwalintukan:
image


liviavinci:
why is this necessary?

dwalintukan:
Oh we talked about it during the meeting. So technically the last voting round (which doesn't hit the threshold) is not validated. Meaning there is no result for that voting round.

So we only want to show results where they are validated: result set round and any voting rounds that hit the threshold.

PR #370, opened 2018-03-22
Elec loading screen

dwalintukan:
Created a simple loading screen for the Electron window. This is separate from the app loading screen, which we cannot show until qtumd is running.

This screen is a loading screen before qtumd is fully running (about 4-5 secs).

image


liviavinci:
@deric the code looks good, but im curious why is the font not Lato?

liviavinci:
Im thinking it might be becuz you didn't do "@import url('/fonts/lato.css');
 " and the style here is separate the app style

PR #378, opened 2018-03-23
Full app navigation

dwalintukan:
- Added back button
- Fixed showing nav arrow in detail pages

image


src/scenes/Activities/index.js
```diff
@@ -84,27 +120,36 @@ class Activities extends React.Component {
   }
 
   handleTabChange(event, value) {
-    this.setState({ tabIdx: value });
+    switch (value) {
+      case 0: {
```
liviavinci: Can we use a const to store the tab index value?

src/scenes/Activities/index.js
```diff
@@ -30,11 +30,47 @@ const messages = defineMessages({
...
+  };
+
   constructor(props) {
     super(props);
 
+    // Determine tab index based on path
+    let tabIdx;
+    switch (this.props.match.path) {
+      case RouterPath.set: {
+        tabIdx = 0;
```
liviavinci: Can we use a const to store the tab index value?

src/scenes/Activities/index.js
dwalintukan: good idea 👍

src/scenes/Activities/index.js
dwalintukan: done

src/components/BackButton/index.jsx
```diff
@@ -0,0 +1,70 @@
...
+
+import styles from './styles';
+import { AppLocation, RouterPath } from '../../constants';
+
+@withRouter
+@injectIntl
+@withStyles(styles, { withTheme: true })
+@connect((state, props) => ({
+  appLocation: state.App.get('appLocation'),
+}), (dispatch, props) => ({
```
alex-cory: @dwalintukan if we're not using the dispatch, we don't need to put it here :)

src/components/NavBar/index.js
```diff
@@ -89,50 +110,54 @@ class NavBar extends React.PureComponent {
...
-
-    return null;
+    return (
+      <img
+        src="/images/nav-arrow.png"
+        alt="nav-arrow"
+        className={
+          classNames(
+            classes.navArrow,
+            currentPath === RouterPath.myWallet || currentPath === RouterPath.set ? 'right' : ''
```
alex-cory: there's a better way to do this. @dwalintukan 
```
classNames(classes.navArrow, {
  right: currentPath === RouterPath.myWallet || currentPath === RouterPath.set
})
```

src/constants.js
```diff
@@ -1,10 +1,33 @@
 module.exports = {
+  RouterPath: {
```
alex-cory: do we use these anywhere else? I personally find it more confusing to have to go to this file and see what is going on than explicitly putting this where it's being used. @dwalintukan

src/components/BackButton/index.jsx
```diff
@@ -0,0 +1,70 @@
...
+}))
+export default class BackButton extends React.Component {
+  static propTypes = {
+    classes: PropTypes.object.isRequired,
+    history: PropTypes.object.isRequired,
+    appLocation: PropTypes.string.isRequired,
+  };
+
+  render() {
+    const {
```
alex-cory: can we please put these on one line unless they're extremely long? Our files just keep getting longer and longer from small things like this <3 @dwalintukan

src/components/BackButton/index.jsx
dwalintukan: gotcha, will remove in a later cleanup PR

dwalintukan: so just dont include the 2nd param?

src/components/BackButton/index.jsx
dwalintukan: sure will remember for the future ones. this one already merged in :P

src/components/NavBar/index.js
dwalintukan: so the { } will return undefined? or will it be `right: undefined`?

PR #380, opened 2018-03-23
add color for disabled button

liviavinci:


liviavinci:
screen shot 2018-03-23 at 9 17 04 pm
screen shot 2018-03-23 at 9 17 15 pm

PR #391, opened 2018-03-24
Prevent user from creating event if they select a creator without enough BOT

liviavinci:


liviavinci:
screen shot 2018-03-25 at 12 21 18 am

src/scenes/CreateEvent/components/CreateEventCreatorPicker/index.js
```diff
@@ -30,18 +41,38 @@ class CreateEventCreatorPicker extends React.PureComponent {
...
+      }
+    </FormControl>);
+  };
+
+  validateEnoughBOT = (address) => {
+    const { walletAddresses, eventEscrowAmount, intl } = this.props;
+
+    for (let i = 0; i < walletAddresses.length; i++) {
+      const checkingAddress = walletAddresses[i];
+      if (checkingAddress.address === address && checkingAddress.bot < eventEscrowAmount) {
```
dwalintukan: use `_.filter(walletAddresses, { address })` to get the walletAddress so you dont have to do a loop.

PR #400, opened 2018-03-26
Implement full withdraw list

liviavinci:
screen shot 2018-03-26 at 10 26 31 pm
screen shot 2018-03-26 at 10 26 38 pm

### DIFF
--- a/src/redux/Topic/reducer.js
+++ b/src/redux/Topic/reducer.js
@@ -1,4 +1,5 @@
 import { Map } from 'immutable';
+import _ from 'lodash';
 import actions from './actions';
 
 const initState = new Map({
@@ -6,6 +7,7 @@ const initState = new Map({
   voteBalances: [],
   botWinnings: 0,
   qtumWinnings: 0,
+  walletAddressesWithWinnings: [],
 });
 
 export default function topicReducer(state = initState, action) {
@@ -29,8 +31,9 @@ export default function topicReducer(state = initState, action) {
         return state.set('errorTopic', action.error);
       }
       return state
-        .set('botWinnings', action.value.botWon)
-        .set('qtumWinnings', action.value.qtumWon);
+        .set('walletAddressesWithWinnings', action.value)
+        .set('botWinnings', _.sumBy(action.value, (wallet) => wallet.botWon ? wallet.botWon : 0))
+        .set('qtumWinnings', _.sumBy(action.value, (wallet) => wallet.qtumWon ? wallet.qtumWon : 0));
     }
     case actions.CLEAR_ERROR_TOPIC: {
       return state.set('errorTopic', undefined);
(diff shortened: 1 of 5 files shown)


src/redux/Topic/saga.js
```diff
@@ -87,28 +87,37 @@ export function* calculateWinningsHandler() {
...
-      const result = yield call(request, Routes.api.winnings, options);
-      let value;
-      if (result) {
-        const botWon = satoshiToDecimal(result['0']);
-        const qtumWon = satoshiToDecimal(result['1']);
-        value = { botWon, qtumWon };
-      } else {
-        value = { botWon: 0, qtumWon: 0 };
+        // return only winning addresses
+        if (botWon || qtumWon) {
```
dwalintukan: is this true if botWon === 0 or qtumWon === 0?

src/redux/Topic/reducer.js
```diff
@@ -1,11 +1,13 @@
 import { Map } from 'immutable';
+import _ from 'lodash';
 import actions from './actions';
 
 const initState = new Map({
   betBalances: [],
   voteBalances: [],
   botWinnings: 0,
   qtumWinnings: 0,
+  walletAddressesWithWinnings: [],
```
dwalintukan: nit: simpler name, `winningAddresses`

src/scenes/Event/scenes/topic.js
```diff
@@ -1,4 +1,4 @@
-/* eslint react/no-array-index-key: 0 */ // Disable "Do not use Array index in keys" for options since they dont have unique identifier
+/* eslint react/no-array-index-key: 0, no-nested-ternary: 0 */ // Disable "Do not use Array index in keys" for options since they dont have unique identifier
```
dwalintukan: nit: is disabling this rule necessary? or more for convenience? suggest we try to follow the preset rules as much as possible.

src/scenes/Event/scenes/topic.js
```diff
@@ -370,23 +422,36 @@ export default class TopicPage extends React.Component {
...
           id="withdrawDetail.alreadyWithdrawn"
           defaultMessage="You have already withdrawn with this address."
         />,
-        warningTypeClass: EventWarningType.Info,
+        warningTypeClass: classes.withdrawn,
+      };
+    }
+
+    // Can withdraw winning
+    const winniningAddresses = _.filter(walletAddressesWithWinnings, {
```
dwalintukan: can use `_.find()` for better efficiency. filter will iterate through the entire list (o(n) time), while find will stop at the first element found.

src/redux/Topic/saga.js
liviavinci: do you mean if botWon === 0 and qtumWon === 0? if botWon === 0 and qtumWon === 0 it will return false.

dwalintukan: k. just wanted to clarify that it checks botWon > 0 || qtumWon > 0 👍

PR #402, opened 2018-03-26
Fix withdraw items

dwalintukan:


src/helpers/utility.js
```diff
@@ -177,3 +177,21 @@ export function getDetailPagePath(oracles) {
...
   return undefined;
 }
+
+/*
+* Filter out unique votes by voter address, topic address, and option index.
+* Used to query against Topics that you can win.
+*/
+export function getUniqueVotes(votes) {
+  const filtered = [];
+  _.each(votes, (vote) => {
```
alex-cory: looks good. Could do this.
```
export function getUniqueVotes(votes) {
  const filtered = [];
  _.each(votes, (vote) => {
    const { voterQAddress, topicAddress, optionIdx } = vote
    if (!_.find(filtered, { voterQAddress, topicAddress, optionIdx })) {
      filtered.push(vote);
    }
  });
  return filtered;
}
```
but not a big deal :)

PR #415, opened 2018-03-27
changed to use moment-duration-format

alex-cory:


src/index.js
```diff
@@ -12,7 +16,10 @@ ReactDOM.render(
 if (module.hot) {
   module.hot.accept('./appProvider.js', () => {
     const NextAppProvider = AppProvider.default;
-    ReactDOM.render(<NextAppProvider />, document.getElementById('root'));
+    render(
+      <NextAppProvider />,
+      document.getElementById('root')
```
dwalintukan: this is for refreshing right?

src/scenes/Event/scenes/oracle.js
```diff
@@ -690,6 +689,7 @@ export default class OraclePage extends React.Component {
     const { oracle } = this.state;
     const totalAmount = _.sum(oracle.amounts);
     const { locale, messages: localeMessages } = this.props.intl;
+    window.oracle = oracle;
```
dwalintukan: is this normal to have the window hold the var?

src/index.js
alex-cory: honestly, I don't think we actually need this. I rewrote all this in my previous company and I removed this and refreshing worked just fine.

src/scenes/Event/scenes/oracle.js
alex-cory: ahhh! Sorry. Totally debugging and forgot to remove. Good catch

src/index.js
dwalintukan: lets leave this to the cleanup after mainnet release. for now, if its not broke, dont fix it.

dwalintukan:
:shipit:

PR #417, opened 2018-03-27
Withdraw escrow

dwalintukan:
image


src/redux/Topic/saga.js
```diff
@@ -116,19 +146,24 @@ export function* calculateWinningsHandler() {
...
         // return only winning addresses
         if (botWon || qtumWon) {
-          value.push({ address: senderAddress, botWon, qtumWon });
+          withdrawableAddresses.push({
+            type: WithdrawType.winnings,
+            address: vote.voterQAddress,
+            botWon,
+            qtumWon,
+          });
         }
```
liviavinci: missing the escrow address i believe?

dwalintukan: it's not needed. the escrow holder is the AddressManager. As long as the BE knows the topicAddress, the topic has the AddressManager's address so it knows where to withdraw from.

src/network/graphMutation.js
```diff
@@ -129,8 +129,9 @@ export function createFinalizeResultTx(version, topicAddress, oracleAddress, sen
   return new GraphMutation('finalizeResult', args, TYPE.transaction).execute();
 }
 
-export function createWithdrawTx(version, topicAddress, senderAddress) {
+export function createWithdrawTx(type, version, topicAddress, senderAddress) {
   const args = {
+    type,
```
alex-cory: you know you can get all the args like this?
```
export const createWithdrawTx = (...args) => new GraphMutation('withdraw', args, TYPE.transaction).execute();
```

alex-cory: jk, they're an array instead of an object then.

dwalintukan: lol

src/scenes/Event/scenes/topic.js
```diff
@@ -360,31 +345,32 @@ export default class TopicPage extends React.Component {
...
-    const config = this.getActionButtonConfig(walletWithinWinnings.address);
+    const config = this.getActionButtonConfig(withdrawableAddress.address);
 
     if (!config.show) {
       return null;
     }
 
     return (
-      <TableRow key={walletWithinWinnings.address}>
+      <TableRow key={index}>
```
liviavinci: Doesn't using index as key give you a warning?

src/scenes/Event/scenes/topic.js
```diff
@@ -360,31 +345,32 @@ export default class TopicPage extends React.Component {
...
         <TableCell padding="dense">
-          <div>{walletWithinWinnings.address}</div>
+          <div>{withdrawableAddress.address}</div>
           <div className={config.warningTypeClass}>{config.message}</div>
         </TableCell>
-        <TableCell padding="dense">Winning</TableCell>
+        <TableCell padding="dense">{withdrawableAddress.type.toUpperCase()}</TableCell>
```
liviavinci: localization?

src/scenes/Event/scenes/topic.js
dwalintukan: nope. it was actually giving me an error if i had an escrow and winnings withdraw. would only show one.

src/scenes/Event/scenes/topic.js
dwalintukan: done

PR #430, opened 2018-03-28
Result history oracle detail

dwalintukan:
image

image

### DIFF
--- a/src/scenes/Event/scenes/oracle.js
+++ b/src/scenes/Event/scenes/oracle.js
@@ -178,7 +178,7 @@ export default class OraclePage extends React.Component {
 
     // Construct page config
     if (oracles) {
-      this.constructOracleAndConfig(syncBlockTime, oracles);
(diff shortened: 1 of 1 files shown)


src/scenes/Event/scenes/oracle.js
```diff
@@ -414,7 +428,7 @@ export default class OraclePage extends React.Component {
     setAppLocation(AppLocation.bet);
 
     return {
-      name: 'BETTING',
+      eventStatus: EventStatus.Bet,
```
alex-cory: Just curious, but what is the purpose of this? I personally find the code harder to read, but I'm sure there's a reason you want to do it.
My guess is...

Pros:
1. if you want to change them all, you can change them in 1 place. (but the variables are named the same thing, i.e. `'BET'` and `EventStatus.Bet` so to utilize this pro you'd have to go change every variable to be `EventStatus.Whatever` anyway everywhere in the codebase.)

Cons:
1. harder to read due to more back and forth between files
2. because of harder to read, slows down development time

Issue #432, opened 2018-03-29
Raised amount doesn't right


henryzbw:
![image](https://user-images.githubusercontent.com/10339131/38063873-7a3b89ee-32b0-11e8-8635-490a62fb549e.png)
![image](https://user-images.githubusercontent.com/10339131/38063884-8a235fee-32b0-11e8-8b12-cf7739731bb3.png)
it has gone through 4 rounds voting, but shows 0 BOT raised

dwalintukan:
The amount that is showing is actually that Oracle's current round. When a `Voting` status Oracle runs out of arbitration time, that same Oracle shows up on the `Finalize` tab. So whatever was bet during the last Voting round would show up there. It's not a total of all the rounds. We actually don't that info in the Oracle.

I will change it so it doesn't show the BOT amount for Finalize cards. This makes more sense since the amount that was voted on that Oracle is inconsequential since it didn't hit the threshold.

PR #435, opened 2018-03-29
Finalize amount

dwalintukan:
#432 
Issue was that the amount did not reflect how much QTUM/BOT was raised in all the rounds. With current BE scheme, we would have to do many hacky calls just to get the total amounts.

Removed `Raised` label for `Finalize` cards. Showing this info is inconsequential for this stage since it doesn't matter how much was raised during this Oracle since it didn't hit the threshold.

image

### DIFF
--- a/src/components/EventCardsGridContainer/index.js
+++ b/src/components/EventCardsGridContainer/index.js
@@ -267,21 +267,18 @@ class EventCardsGrid extends React.Component {
     _.each(oracles, (oracle) => {
       const amount = parseFloat(_.sum(oracle.amounts).toFixed(2));
       let buttonText;
-      let amountLabel;
+      let amountLabel = `${amount} ${oracle.token}`;
       switch (eventStatusIndex) {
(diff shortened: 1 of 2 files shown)


src/components/EventCardsGridContainer/index.js
```diff
@@ -265,39 +265,43 @@ class EventCardsGrid extends React.Component {
   renderOracles(oracles, eventStatusIndex) {
     const rowItems = [];
     _.each(oracles, (oracle) => {
+      const amount = parseFloat(_.sum(oracle.amounts).toFixed(2));
       let buttonText;
+      let amountLabel;
```
alex-cory: we could set the default to 
```
let amountLabel = `${amount} ${oracle.token}`;
```
and only in the Finalize set it to undefined. Save 3 lines :)

dwalintukan: good call, fixed 👍

PR #437, opened 2018-03-29
Finalize mutation

dwalintukan:
Pass optionIdx to finalize result mutation so we can see it in the tx history.

image

Depends on: https://github.com/bodhiproject/bodhi-graphql/pull/172


dwalintukan:
not needed, logic on BE now.

PR #453, opened 2018-03-30
Tx history pagination

dwalintukan:
image

image


src/network/graphQuery.js
```diff
@@ -186,14 +186,20 @@ export function queryAllVotes(filters, orderBy) {
...
-export function queryAllTransactions(filters, orderBy) {
+export function queryAllTransactions(filters, orderBy, limit, skip) {
   const request = new GraphQuery('allTransactions', TYPE.transaction);
   if (!_.isEmpty(filters)) {
     request.setFilters(filters);
   }
   if (!_.isEmpty(orderBy)) {
     request.setOrderBy(orderBy);
   }
+  if (_.isFinite(limit) && limit > 0) {
```
liviavinci: why is finite not !isEmpty?

dwalintukan: `isFinite` is checking if it is a finite number (non-theoretical) and also for undefined too. as most lodash checks do.

`limit` is a number so should be an undefined or > check.

PR #458, opened 2018-03-31
New design for loading with more info

liviavinci:


liviavinci:
screen shot 2018-03-31 at 7 21 38 pm

liviavinci:
@dwalintukan i realized that since we changed the default locale to Chinese, all first time users will be seeing the loader in Chinese...

PR #459, opened 2018-04-01
Withdraw calculations

dwalintukan:


src/languageProvider/locales/zh-Hans.json
```diff
@@ -185,11 +185,7 @@
   "withdrawDetail.alreadyWithdrawn": "您已经使用过此地址提现。",
   "withdrawDetail.returnRate": "回报率：",
   "withdrawDetail.reward": "回报",
-  "withdrawDetail.totalBet": "总计投注",
```
liviavinci: can you fill in a temp str for the new vars?

dwalintukan: ran the build langs script

PR #472, opened 2018-04-04
Prevent showing empty percent or really old date

liviavinci:


src/scenes/App/components/Loader/index.js
```diff
@@ -49,20 +50,22 @@ class Loader extends React.PureComponent {
...
-              {syncBlockNum}
-            </Grid>
-            <Grid item className={classes.loaderInfoLabel} xs={6}>
-              <FormattedMessage id="loader.blockTime" defaultMessage="Latest Block Time" />
-            </Grid>
-            <Grid item className={classes.loaderInfoData} xs={6}>
-              {getLocalDateTimeString(syncBlockTime)}
-            </Grid>
-          </Grid>
+          { !_.isEmpty(syncBlockNum) && !_.isEmpty(syncBlockTime) ?
```
dwalintukan: should't it be `syncBlockNum && syncBlockTime`? according to lodash docs (https://lodash.com/docs/4.17.5#isEmpty), don't think isEmpty handles numbers correctly.

Issue #477, opened 2018-04-04
user's default language not sticking


alex-cory:
When I switch to english, it always switches back to mandarin when I refresh

frankobe:
@alex-cory is this fixed?

alex-cory:
@frankobe yep!

PR #478, opened 2018-04-04
changed the way we are doing the language switch

alex-cory:
so its more intuitive to the developer and also fixed it


alex-cory:
https://github.com/bodhiproject/bodhi-ui/issues/477

PR #482, opened 2018-04-06
route change for qtum mainnet

alex-cory:


src/network/routes.js
```diff
@@ -2,11 +2,13 @@ const AUTHORITY = '127.0.0.1:5555';
 const HTTP_ROUTE = `http://${AUTHORITY}`;
 const WS_ROUTE = `ws://${AUTHORITY}`;
 
-const QTUM_EXPLORER_DEV = 'https://testnet.qtum.org';
-const QTUM_EXPLORER_PROD = 'https://explorer.qtum.org';
 
-const BASE_INSIGHT_DEV = `${QTUM_EXPLORER_DEV}/insight-api`;
-const BASE_INSIGHT_PROD = `${QTUM_EXPLORER_PROD}/insight-api`;
+let QTUM_EXPLORER = 'https://explorer.qtum.org';
+let BASE_INSIGHT = `${QTUM_EXPLORER}/insight-api`;
```
dwalintukan: move this under the if block, then remove the `BASE_INSIGHT` in the if block so only need to set it once

src/network/routes.js
```diff
@@ -2,11 +2,13 @@ const AUTHORITY = '127.0.0.1:5555';
...
+if (process.env.REACT_APP_ENV === 'dev') {
+  QTUM_EXPLORER = 'https://testnet.qtum.org';
+  BASE_INSIGHT = `${QTUM_EXPLORER}/insight-api`;
```
dwalintukan: remove after moving BASE_INSIGHT

PR #486, opened 2018-04-09
Initializing chinese

liviavinci:


liviavinci:
@frankobe @henryzbw since i didn't find a way to do localization in html, just used a Chinese string. This should be the same as doing localization anyway since in the init stage user cannot change their locale.

src/html/loading/index.html
```diff
@@ -7,7 +7,7 @@
   <body>
     <div class="container">
       <div class="loader"></div>
-      <h1>Initializing...</h1>
+      <h1>初始化中...</h1>
```
frankobe: @liviavinci This wont solve the problem as English user will see the page again and again. HTML localization usually takes places during rendering. As this page is load directly from backend (which doesnt control localization now), I can come up with two potential solutions:

1. make this string into a React component and localized from React
2. remove this string completely

dwalintukan: Regarding #1, this is not possible (AFAIK) since this html is not part of the react code at all. It is independent entirely.

alex-cory: Why don't we just put in some JS like
```
<h1 id='loading-text'>初始化中...</h1>
<script>
  const locale = window.navigator.userLanguage || window.navigator.language;
  const loadingText = {
    'en-US': 'Initializing...',
    'zh-Hans-CN': '初始化中...',
  }
  document.getElementById('main-loader').innerHTML = loadingText[locale]
</script>
```

frankobe: @alex-cory I am fine with removing the text as loading animation means `intializing...`

src/html/loading/index.html
```diff
@@ -7,7 +7,12 @@
   <body>
     <div class="container">
       <div class="loader"></div>
-      <h1>Initializing...</h1>
+      <h1 id='loading-text'>初始化中...</h1>
+      <script>
+        const locale = window.navigator.userLanguage || window.navigator.language;
+        const loadingText = local.startsWithg('en') ? 'Initializing...' : '初始化中...';
```
henryzbw: `local.startsWithg('en')`
should be
`locale.startsWith('en')`

src/html/loading/index.html
```diff
@@ -7,7 +7,12 @@
   <body>
     <div class="container">
       <div class="loader"></div>
-      <h1>Initializing...</h1>
+      <h1 id='loading-text'>初始化中...</h1>
+      <script>
+        const locale = window.navigator.userLanguage || window.navigator.language;
```
henryzbw: `local.startsWithg('en')`
should be
`locale.startsWith('en')`

src/html/loading/index.html
```diff
@@ -7,7 +7,12 @@
...
+        const loadingText = local.startsWith('en') ? 'Initializing...' : '初始化中...';
+        document.getElementById('loading-text').innerHTML = lodingText;
```
frankobe: `lodingText` - >`loadingText `

src/html/loading/index.html
frankobe: @alex-cory will locale be `undefined`? if so, L13 will throw an error. If not, LGTM

alex-cory: no those 2 cases should cover all browsers

PR #488, opened 2018-04-09
set default language based on browser language

henryzbw:
detect browser language first, then check localstorage, if localstorage doesn't exist, set the default language by browser language, if the user changed the language once, it will set localstorage,  the new language will become the default language next time he starts the application.

### DIFF
--- a/src/appProvider.js
+++ b/src/appProvider.js
@@ -11,13 +11,20 @@ import AppLocale from './languageProvider';
 import bodhiTheme from './config/theme';
 import graphClient from './network/graphClient';
 import { store, history } from './redux/store';
-
 import '../src/style/styles.less';
+
+const language = (navigator.languages && navigator.languages[0]) ||
+                     navigator.language ||
+                     navigator.userLanguage;
+let browserLang = 'mandarin';
+if (language.includes('en')) {
+  browserLang = 'english';
+}
 
 export default class AppProvider extends Component {
   locales = { english: AppLocale.en, mandarin: AppLocale.zh }
   state = {
-    locale: localStorage.getItem('language') || 'mandarin',
+    locale: localStorage.getItem('language') || browserLang,
   }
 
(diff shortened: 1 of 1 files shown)


src/appProvider.js
```diff
@@ -11,13 +11,20 @@ import AppLocale from './languageProvider';
 import bodhiTheme from './config/theme';
 import graphClient from './network/graphClient';
 import { store, history } from './redux/store';
-
 import '../src/style/styles.less';
 
+const language = (navigator.languages && navigator.languages[0]) ||
```
frankobe: consider match elements in `navigator.languages` in order.

src/appProvider.js
```diff
@@ -11,13 +11,20 @@ import AppLocale from './languageProvider';
...
+                     navigator.language ||
+                     navigator.userLanguage;
+let browserLang = 'mandarin';
```
frankobe: why not using `AppLocale.zh`?

src/appProvider.js
```diff
@@ -11,13 +11,20 @@ import AppLocale from './languageProvider';
...
+if (language.includes('en')) {
```
frankobe: Need ensure language starts with `en` or `en-`.  `en` can be in the middle of the string

src/appProvider.js
alex-cory: readability. Since we're only using 2 languages right now, I implemented it this way

frankobe: @alex-cory dunt think readability is an issue, `cn` and `en` are commonly used

src/appProvider.js
alex-cory: this code block should go in `componentDidMount`

src/appProvider.js
alex-cory: yeah I worked with him to get a better solution :)

PR #499, opened 2018-04-16
Fix wallet unlock for encrypted wallets

dwalintukan:
image

image


src/components/WalletUnlockDialog/index.jsx
```diff
@@ -23,22 +23,30 @@ const messages = defineMessages({
...
-  static defaultProps = {
-  };
-
   constructor(props) {
     super(props);
 
     this.state = {
-      unlockMinutes: 1,
+      passphrase: '',
+      unlockMinutes: 5,
```
liviavinci: let's use a const for the default value?

dwalintukan: added a default unlockWalletMins in `config`

PR #507, opened 2018-04-19
add important note in wallet page

### LINKED ISSUE #505: Add info text for launching qtum wallet
image

Add this icon and text to the `My Wallet` page at the very top:
```
If you would like to encrypt, backup, or restore your wallet, please do this from the Qtum Wallet. In the main top menu, there is a "Launch Qtum Wallet" option.
```

Add this above the My Balance section.
image

Also this needs to be localized.

henryzbw:
image
fix  #505

### DIFF
--- a/src/scenes/Wallet/components/Balances/index.jsx
+++ b/src/scenes/Wallet/components/Balances/index.jsx
@@ -13,7 +13,7 @@ import CloseIcon from 'material-ui-icons/Close';
 import ContentCopy from 'material-ui-icons/ContentCopy';
 import { withStyles } from 'material-ui/styles';
 import classNames from 'classnames';
-import { FormattedMessage, injectIntl, defineMessages } from 'react-intl';
+import { FormattedMessage, injectIntl } from 'react-intl';
 import { CopyToClipboard } from 'react-copy-to-clipboard';
 import _ from 'lodash';
 
--- a/src/scenes/Wallet/components/Balances/index.jsx
+++ b/src/scenes/Wallet/components/Balances/index.jsx
@@ -13,10 +13,9 @@ import CloseIcon from 'material-ui-icons/Close';
 import ContentCopy from 'material-ui-icons/ContentCopy';
 import { withStyles } from 'material-ui/styles';
 import classNames from 'classnames';
-import { FormattedMessage, injectIntl, defineMessages, formatMessage, intlShape } from 'react-intl';
+import { FormattedMessage, injectIntl, defineMessages } from 'react-intl';
 import { CopyToClipboard } from 'react-copy-to-clipboard';
 import _ from 'lodash';
-import ImportantNote from '../../../../components/ImportantNote/index';
 
 import styles from './styles';
 import Config from '../../../../config/app';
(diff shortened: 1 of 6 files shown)


src/scenes/Wallet/components/Balances/index.jsx
```diff
@@ -70,6 +78,11 @@ class MyBalances extends React.PureComponent {
       <Paper className={classes.myBalancePaper}>
         <Grid container spacing={0} className={classes.myBalanceGridContainer}>
           <Typography variant="title" className={classes.myBalanceTitle}>
+            <div className={classes.importantNoteContainer}>
```
alex-cory: instead of adding an extra wrapper `<div>`, why don't you just add `margin-bottom` in a class to `<ImportantNote />`?
You'll have to extend `ImportantNote` by putting `<div {...props}>` in it

alex-cory: I would also say add the `injectIntl` in `ImportantNote`, but I know how much they don't want refactoring right now so we can hold off on that for later unless @dwalintukan says it's cool. It's only in 3 places in the codebase

dwalintukan: move this `<div>` to `src/scenes/Wallet/index.jsx`. this does not belong in the MyBalances component.

dwalintukan: let's keep it the way it is right now. for certain important note messages, you would have to pass in the string to translate as well as some vars that go into that string.

dwalintukan: also wrap it in a `<Paper>` component when moving it. and move the marginBottom. so it looks like a separate piece of paper on the page.

src/scenes/Wallet/components/Balances/index.jsx
```diff
@@ -13,7 +13,7 @@ import CloseIcon from 'material-ui-icons/Close';
 import ContentCopy from 'material-ui-icons/ContentCopy';
 import { withStyles } from 'material-ui/styles';
 import classNames from 'classnames';
-import { FormattedMessage, injectIntl } from 'react-intl';
+import { FormattedMessage, injectIntl, defineMessages } from 'react-intl';
```
dwalintukan: remove defineMessages

src/scenes/Wallet/index.jsx
```diff
@@ -19,8 +34,15 @@ class MyWallet extends React.Component {
   }
 
   render() {
+    const { classes } = this.props;
+    console.log(classes);
```
dwalintukan: remove

src/scenes/Wallet/styles.js
```diff
@@ -0,0 +1,14 @@
+const styles = (theme) => ({
+  infoPaper: {
+    minWidth: theme.sizes.table.minWidth,
+    overflowX: 'scroll',
+    boxShadow: 'none',
+    borderRadius: theme.borderRadius,
+    background: theme.palette.background.paper,
+    paddingTop: '10px',
+    paddingLeft: theme.padding.md.px,
+    marginBottom: '10px',
```
dwalintukan: theme.padding.xs.px

src/scenes/Wallet/styles.js
```diff
@@ -0,0 +1,14 @@
+const styles = (theme) => ({
+  infoPaper: {
+    minWidth: theme.sizes.table.minWidth,
+    overflowX: 'scroll',
+    boxShadow: 'none',
+    borderRadius: theme.borderRadius,
+    background: theme.palette.background.paper,
+    paddingTop: '10px',
+    paddingLeft: theme.padding.md.px,
```
dwalintukan: when setting padding, its best to make it symmetrical. so top/bottom and left/right should be the same. you can do it like this:

```
padding: `${theme.padding.xs.px} ${theme.padding.md.px}`
```

The first value is the top/bottom, and the second value is left/right.

PR #508, opened 2018-04-19
Dashboard test

alex-cory:
- setup the app for testing (i.e. `createStore`, making apollo/websockets compatible)
- created 1st tests
- moved `global.store` to `dev` env only
- created mock client for testing
- dashboard tests
- graphql tests (WIP)


src/redux/Dashboard/reducer.js
```diff
@@ -1,20 +1,25 @@
...
 export default function dashboardReducer(state = initState, action) {
   switch (action.type) {
-    case actions.TAB_INDEX_CHANGED: {
-      return state.set('tabIndex', action.value);
-    }
-    case actions.SORT_ORDER_CHANGED: {
-      return state.set('sortBy', action.sortBy);
-    }
-    default: {
+    case actions.TAB_INDEX_CHANGED:
```
alex-cory: I noticed this is dead code. Before merging this, unless theirs any objections, I'm going to remove this

src/redux/Dashboard/reducer.js
```diff
@@ -1,20 +1,25 @@
 import { Map } from 'immutable';
 import actions from './actions';
+import { SortBy } from '../../constants';
 
-const initState = new Map({
+const initState = {
   value: undefined,
```
alex-cory: Pretty sure this isn't being used either

src/redux/Dashboard/reducer.js
dwalintukan: yup remove anything not being used.

src/network/graphClient.js
```diff
@@ -11,7 +11,9 @@ const httpLink = new HttpLink({
   uri: Routes.graphql.http,
 });
 
-const wsLink = new WebSocketLink({
+// the check for process.browser makes sure our tests run properly since
+// this can only be run in the browser. Source: http://tinyurl.com/ycsjfq6p
+const wsLink = process.browser && new WebSocketLink({
```
dwalintukan: according to the source you posted, it should be `process.browser ? ...`

dwalintukan: can you explain what the issue of having this check is for in relation to testing? if this is just for the dashboard tests, why would the graphClient need to be changed? the sortOrder test does not use the graphClient?

src/redux/reducers.js
```diff
@@ -1,11 +1,11 @@
 import App from './App/reducer';
-import Dashboard from './Dashboard/reducer';
+import dashboard from './Dashboard/reducer';
```
dwalintukan: is this proper naming convention for lowercase?

src/redux/store.js
```diff
@@ -17,14 +17,23 @@ if (process.env.REACT_APP_ENV === 'dev') {
...
+    compose(applyMiddleware(...middlewares))
+  );
+  sagaMiddleware.run(rootSaga);
+  return store;
+};
+
+const store = createStore();
+
+if (process.env.REACT_APP_ENV === 'dev') {
+  window.store = store;
```
dwalintukan: this does not need to be in prod env? please test with a prod build with Electron.

src/redux/reducers.js
alex-cory: yep

alex-cory: in majority of apps I've read and written

src/redux/store.js
alex-cory: this does not. I added this for testing purposes only so you could check the store's state in the console

src/network/graphClient.js
alex-cory: 1. when you're using a ternary, in most cases, if you're returning `null`, it's not necessary to use a ternary. The `&&` does essentially the same thing.
```
const wsLink = process.browser && whatever // "undefined" if no in browser
const wsLink = process.browser ? whatever : null // "null" if no browser
```

2. I'm honestly not quite sure why this is happening. From what I read this seems to be a SSR issue (maybe because it's a test? idk). I tried this solution out and it worked. The error I was getting was:
image
I found [this issue](https://github.com/apollographql/subscriptions-transport-ws/issues/87) that's pretty similar, but the implementations they're using are different. I can do some more digging, but I have a feeling I'd have to change our implementation/it might take longer than desired to do this a different way.

alex-cory: Basically the library isn't getting the `WebSocket` library it needs for some reason. I saw some other people say they tried making it `global`, but that still didn't work.

dwalintukan: 1. 👍 
2. My main concern is that prod Electron version might be different for `process.browser` (may be undefined). So if `process.browser` is undefined in the Electron prod version the WS won't work. Would like to hold merging this PR until after mainnet release. Code looks fine to me, but just want to make sure nothing breaks right before our launch.

src/redux/reducers.js
dwalintukan: 👍

src/redux/store.js
dwalintukan: gotcha, so you added the window.store = store before.

dwalintukan:
Approved but please hold off on merging until after mainnet release. See comment above ^

src/network/graphClient.js
dwalintukan: Electron is not your typical environment. Like for example, `process.env` was not able to be found in the prod version. Something with how Electron's main/renderer processes work.

src/redux/store.js
alex-cory: correct

src/network/graphClient.js
alex-cory: makes total sense. I'll check that out tomorrow

amazingandyyy:
close it due to inactivities, can open it again if you wish.

PR #518, opened 2018-04-25
Loading fix

alex-cory:
adding padding


src/scenes/App/components/Loader/styles.js
```diff
@@ -63,9 +63,11 @@ const styles = (theme) => ({
   loaderInfoLabel: {
     fontWeight: 'bold',
     textAlign: 'right',
+    padding: 6,
```
dwalintukan: use theme.padding values?

PR #543, opened 2018-05-08
make card height adjustable

liviavinci:
screen shot 2018-05-08 at 12 57 55 pm


dwalintukan:
@liviavinci what do you think about just ellipsizing the title if it gets too big?

liviavinci:
screen shot 2018-05-08 at 5 41 36 pm

PR #545, opened 2018-05-08
Fix Javascript float subtraction problem

liviavinci:
screen shot 2018-05-08 at 9 46 44 pm


src/scenes/Event/scenes/oracle.js
```diff
@@ -663,6 +664,25 @@ export default class OraclePage extends Component {
     };
   }
 
+  toFixed = (num) => {
```
dwalintukan: we cant use the regular `number.toFixed()`?

dwalintukan: if not, lets move this into a util/helper class.

PR #551, opened 2018-05-10
Fix weird zero string

liviavinci:


liviavinci:
Before:
screen shot 2018-05-10 at 1 11 28 pm
After:
screen shot 2018-05-10 at 2 14 37 pm

PR #563, opened 2018-05-22
fix pagination

henryzbw:
add more initial oracles so we can touch the bottom of screen to trigger the loadmore function.


src/components/EventCardsGridContainer/index.js
```diff
@@ -39,7 +39,7 @@ const messages = defineMessages({
 });
 
 const { Pending } = TransactionStatus;
-const LIMIT = 8;
+const LIMIT = 16;
```
dwalintukan: lets change this to a higher number so it covers all screen resolutions. lets do 50

PR #574, opened 2018-05-25
quick clean

alex-cory:
- combined imports
- removed redundant method
- removed unused variables
- added keys to list
- removed console warnings/errors


src/components/TxConfirmDialog/index.js
```diff
@@ -104,16 +97,9 @@ export default class TxConfirmDialog extends Component {
...
-  }
-
   onOkClicked = () => {
-    const callback = this.props.txConfirmInfoAndCallback.confirmCallback;
-    if (callback) {
-      callback();
-    }
-    this.props.clearTxConfirm();
+    const { txConfirmInfoAndCallback: { confirmCallback }, clearTxConfirm } = this.props;
+    if (confirmCallback) clearTxConfirm();
```
dwalintukan: i believe we still need to call `confirmCallback` as the function as well?

alex-cory: good catch

alex-cory: fixed

PR #576, opened 2018-05-29
adding routes to the graphql saga

alex-cory:


src/redux/Graphql/saga.js
```diff
@@ -45,13 +46,17 @@ export function* getTopicsHandler() {
...
-    } catch (err) {
-      console.error(err); // eslint-disable-line
+    } catch (error) {
+      console.error(error); // eslint-disable-line
       yield put({
         type: actions.GET_TOPICS_RETURN,
         value: [],
         limit: action.limit,
         skip: action.skip,
+        error: {
```
dwalintukan: remove this error obj.

reason:
let's not throw an error for graphql calls that are called in the background (only for ones that require user interaction ie. button click). 

this would pop up every time if they went to the dashboard and could be really annoying if they have an error. graphql calls that call in the bg should fail silently and we can catch this error with remote error logging on UI side.

src/redux/Graphql/saga.js
```diff
@@ -87,13 +92,17 @@ export function* getActionableTopicsHandler() {
...
-    } catch (err) {
-      console.log(err); // eslint-disable-line
+    } catch (error) {
+      console.log(error); // eslint-disable-line
       yield put({
         type: actions.GET_TOPICS_RETURN,
         value: [],
         limit: action.limit,
         skip: action.skip,
+        error: {
```
dwalintukan: remove error obj. see comment above.

src/redux/Graphql/saga.js
```diff
@@ -123,13 +132,17 @@ export function* getOraclesHandler() {
...
-    } catch (err) {
-      console.error(err); // eslint-disable-line
+    } catch (error) {
+      console.error(error); // eslint-disable-line
       yield put({
         type: actions.GET_ORACLES_RETURN,
         value: [],
         limit: action.limit,
         skip: action.skip,
+        error: {
```
dwalintukan: remove error obj. see comment above

src/redux/Graphql/saga.js
```diff
@@ -148,11 +161,15 @@ export function* getTransactionsHandler() {
...
         skip: action.skip,
       });
-    } catch (err) {
-      console.error(err); // eslint-disable-line
+    } catch (error) {
+      console.error(error); // eslint-disable-line
       yield put({
         type: actions.GET_TRANSACTIONS_RETURN,
         value: [],
+        error: {
```
dwalintukan: remove error obj. see comment above

src/redux/Graphql/saga.js
```diff
@@ -186,11 +203,15 @@ export function* getPendingTransactionsHandler() {
...
         value: pendingTxsObj,
       });
-    } catch (err) {
-      console.error(err); // eslint-disable-line
+    } catch (error) {
+      console.error(error); // eslint-disable-line
       yield put({
         type: actions.GET_PENDING_TRANSACTIONS_RETURN,
         value: [],
+        error: {
```
dwalintukan: remove error obj. see comment above

dwalintukan:
:shipit:

Issue #594, opened 2018-06-05
Detail page does not refresh after doing a tx


dwalintukan:
1. Go to my wallet page
2. Withdraw qtum/bot
3. Transfer history does not refresh

This also happens on the Oracle/Topic detail pages.
1. go to event detail
2. do tx (eg. bet)
3. the Transaction History does not get updated. have to go out of the page and come back in to see it.

dwalintukan:
#606 fixed

Issue #604, opened 2018-06-07
Create an event button doesn't work


henryzbw:
Just found the `create an event` button doesnt work

dwalintukan:
#606 fixed

PR #606, opened 2018-06-07
Use undefined for txReturn instead of empty obj

dwalintukan:
This fixes:
https://github.com/bodhiproject/bodhi-ui/issues/594
https://github.com/bodhiproject/bodhi-ui/issues/604

### DIFF
--- a/src/components/TransactionSentDialog/index.js
+++ b/src/components/TransactionSentDialog/index.js
@@ -4,7 +4,6 @@ import { connect } from 'react-redux';
 import { Button, Typography, withStyles } from 'material-ui';
 import Dialog, { DialogActions, DialogContent, DialogTitle } from 'material-ui/Dialog';
 import { injectIntl, intlShape, defineMessages } from 'react-intl';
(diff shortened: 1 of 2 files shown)


src/components/TransactionSentDialog/index.js
```diff
@@ -45,14 +44,15 @@ export default class TransactionSentDialog extends Component {
   }
 
   static defaultProps = {
-    txReturn: {},
+    txReturn: undefined,
   }
 
   render() {
     const { intl, classes, txReturn, clearTxReturn } = this.props;
+    if (!txReturn) return null;
```
alex-cory: This is the line that needs to get removed so that when a transaction happens, this dialog will display. Without this line, line 59 will error. `cannot access txid of undefined` because it's not an object.
We can do a quick fix to this by just making the default in the component be `{}`, and the global one be `undefined`

dwalintukan: tested already with multiple txs and is working.

PR #607, opened 2018-06-07
fix after creation, it wont show immediately

### LINKED ISSUE #479: when on Oracle page after creating event and waiting on next block, page goes blank
I think this happens when the next block comes in because the `topicAddress` and the `address` in the url are `null` and we use those in the constructor to get the `oracle` and set it to state. Since their's no oracle, nothing get's rendered. See these lines

image

Here's the url while waiting on the event to confirm on the blockchain.
`http://localhost:3000/oracle/null/null/c5a2caed2c06b4eb3b839d2e9bbaf749ec37f51065ee1e5c18ab5f4c7f0781fb`

Maybe we need to force a refresh when these values come in or better yet, don't put them in the url then we don't have to force a refresh.

image

henryzbw:
fix #479

### DIFF
--- a/src/components/EventCardsGridContainer/index.js
+++ b/src/components/EventCardsGridContainer/index.js
@@ -51,6 +51,7 @@ const SKIP = 0;
   sortBy: state.Dashboard.get('sortBy'),
   syncBlockNum: state.App.get('syncBlockNum'),
   walletAddresses: state.App.get('walletAddresses'),
+  txReturn: state.Graphql.get('txReturn'),
 }), (dispatch) => ({
   setAppLocation: (location) => dispatch(appActions.setAppLocation(location)),
   getActionableTopics: (walletAddresses, orderBy, limit, skip) =>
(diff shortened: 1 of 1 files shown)


src/components/EventCardsGridContainer/index.js
```diff
@@ -71,9 +72,11 @@ export default class EventCardsGrid extends Component {
     walletAddresses: PropTypes.array.isRequired,
     intl: intlShape.isRequired, // eslint-disable-line react/no-typos
     status: PropTypes.string,
+    txReturn: PropTypes.object,
   };
 
   static defaultProps = {
+    txReturn: undefined,
```
dwalintukan: this should be unnecessary since the propType is not `.isRequired`

henryzbw: @dwalintukan make txReturn required will throw warning ```Warning: Failed prop type: The prop `txReturn` is marked as required in `EventCardsGrid`, but its value is `undefined`.```

dwalintukan: oh i see because it is not defined in the redux store first

PR #612, opened 2018-06-08
use ternary instead of if else

henryzbw:


src/scenes/Event/components/EventTxHistory/index.js
```diff
@@ -110,11 +110,10 @@ export default class EventTxHistory extends Component {
       case TransactionType.ApproveVote:
       case TransactionType.Vote:
       case TransactionType.FinalizeResult: {
-        if (optionIdx && topic) {
-          const invalidOption = localizeInvalidOption(tx.topic.options[tx.optionIdx], intl);
-          return `#${tx.optionIdx + 1} ${tx.topic.options[tx.optionIdx] === 'Invalid' ? invalidOption : tx.topic.options[tx.optionIdx]}`;
-        }
-        return '';
+        const invalidOption = localizeInvalidOption(tx.topic.options[tx.optionIdx], intl);
```
dwalintukan: won't this throw an error if tx.topic is `undefined`? looks like that was what the safety `if` check was for

henryzbw: yeah, i c. i didn't realize it when i made it.

PR #617, opened 2018-06-11
Fix 0betted show100

### LINKED ISSUE #609: Betting round always show 100 QTUM even 0 qtum betted
image

henryzbw:
fix #609

### DIFF
--- a/src/scenes/Event/components/EventTxHistory/resultHistory.js
+++ b/src/scenes/Event/components/EventTxHistory/resultHistory.js
@@ -36,7 +36,9 @@ export default class EventResultHistory extends Component {
     if (sortedOracles.length) {
       const { resultIdx, options, amounts, consensusThreshold } = sortedOracles[0];
       const { endTime, token } = sortedOracles[1];
-      const resultSettingRound = { endTime, token, resultIdx, options, amounts };
+      const resultSettingRound = { endTime, token, resultIdx, options };
+      const resultSettingRoundamount = amounts.slice(0);
+      resultSettingRound.amounts = resultSettingRoundamount;
       resultSettingRound.amounts.fill(0);
       resultSettingRound.amounts[resultSettingRound.resultIdx] = consensusThreshold;
       sortedOracles.splice(1, 0, resultSettingRound);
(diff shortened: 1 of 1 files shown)


src/scenes/Event/components/EventTxHistory/resultHistory.js
```diff
@@ -36,7 +36,8 @@ export default class EventResultHistory extends Component {
     if (sortedOracles.length) {
       const { resultIdx, options, amounts, consensusThreshold } = sortedOracles[0];
       const { endTime, token } = sortedOracles[1];
-      const resultSettingRound = { endTime, token, resultIdx, options, amounts };
+      const resultSettingRound = { endTime, token, resultIdx, options };
+      resultSettingRound.amounts = amounts.slice(0);
```
dwalintukan: curious why this works? slice(0) is the same thing as copying the array.
`resultSettingRound.amounts = amounts` should be the same outcome?

dwalintukan: i think the issue is that `resultSettingRound.amounts` was referencing the `sortedOracles[0].amounts`. you just need to `copy` the array instead of referencing it.

the `slice(0)` works because it is creating a new array: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice

henryzbw: yes. before i just referred the array, but what i need here is to copy it.

dwalintukan: Yup so use a proper lodash method to copy the array

henryzbw: why need to use lodash to copy? slice works, too

dwalintukan: @henryzbw its like you are using a workaround for the fix when they have a function that does what you want. 

would you eat a steak with a fork and spoon? you could, but the fork and knife are specific for that purpose.

Issue #357, opened 2018-03-20
Find a better date picker and time picker


liviavinci:

lillland:
I found these link: 
http://t00rk.github.io/bootstrap-material-datetimepicker/  
you can pick year/month/day/hour/minute

https://mdbootstrap.com/javascript/date-picker/#events 
this one likes better, but no time picker.

liviavinci:
@lillland the first one is great!

alex-cory:
I think [this one](https://material-ui-pickers.firebaseapp.com/demo/datetimepicker) is the exact one we were looking for.
![image](https://user-images.githubusercontent.com/5455859/40740715-6cc4a8f2-63fe-11e8-99e6-54fa78e96465.png)

amazingandyyy:
in order to add this package, new picker, we need to use @material-ui/core(v1.4) package, current we are using material-ui(v0.9)

In order to use @material-ui/core package, we need to update to React 16

amazingandyyy:
@dwalintukan I have successfully updated to React 16 without error, 
1. not sure if this should be seperate PR
2. not sure if no error means "No side effect"

I also update tried to update to @material-ui/core, which updated over 80 files, will the PR be too fat?
I stay the latest code [here on my own branch](https://github.com/amazingandyyy/bodhi-ui/tree/new_date_picker)

Will need to confirm 
1. do we want to update to React 16?
2. do we want to update to @material-ui/core with many refactor of files?
3. should they be seperate PRs? 

Thanks

dwalintukan:
1. Yes we do want to keep updated. How long is ETA for the refactor?
2/3. If the refactor is going to be big, lets create a master sub-branch then do PRs into that master sub.

amazingandyyy:
ETA for @material-ui/core is around one afternoon, refactor + debug of some incompatible component e.g. Button.

amazingandyyy:
will do separate PR for updating React and breaking @material-ui into multiple PRs

dwalintukan:
@amazingandyyy sounds good. the material UI lib might need a lot of updating too?

amazingandyyy:
Yes! And I original planned to refactor the material ui this afternoon, but I decided to move to tmr’s todo list since going through qweb3 and bodhi-server took more time than I expected. 

Will make the refactor PR send tmr!

ippudo:
I think I may take this task

dwalintukan:
[...]

PR #620, opened 2018-06-12
Upgrade pacakge version

amazingandyyy:
1. fix the issue of the unnecessary `yarn upgrade` everytime of cloning down, by updating yarn-lock and package-lock
2. use `npm audit fix` to safely upgrade pacakges which have security concern to the new package which is guaranteed backward compatible.
3. update package name from `isomorphic-boilerplate` to `bodhi-ui`
4. update `test:eslint` script and `.eslintignore`, to prepare for the proposal of **eslint checking**(by travis) of every PR
4. update script of `test:eslint` and add two script to make sure developer follow the style rules
- `precommit` and `prepush` to run `eslint:test` before the developer can commit or push


amazingandyyy:
I will update React and material-ui(refer to #357 ) after this get merge otherwise to avoid the potential huge conflicts of -lock file.

package.json
```diff
@@ -125,10 +125,12 @@
...
     "build": "node scripts/build.js",
     "test": "node scripts/test.js --env=jsdom",
-    "lint:fix": "eslint --fix --",
+    "lint:fix": "eslint . --fix",
     "extract-messages": "NODE_ENV=production babel ./src  --out-file /dev/null",
     "manage:translations": "NODE_ENV=production ./node_modules/.bin/babel-node scripts/translator.js",
-    "build:langs": "npm run extract-messages && npm run manage:translations"
+    "build:langs": "npm run extract-messages && npm run manage:translations",
+    "precommmit": "npm run lint:fix",
+    "prepush": "npm run lint:fix"
```
dwalintukan: why not just have travis run `lint:fix`?

public/iconfont/iconfont.js
```diff
@@ -1 +1,12 @@
...
\ No newline at end of file
+(function (window) {
```
dwalintukan: looks like this still got linted?

.eslintignore
```diff
@@ -1 +1,4 @@
 src/modules/**
+config/**
+scripts/**
+public/**
```
dwalintukan: add end of line

package.json
dwalintukan: why not just have travis run `lint:fix`?

amazingandyyy: I prefer do both travis and precommit and prepush, the benefit of using precommit and prepush checking is to make the debug lifecycle shorter. For exmaple, if the style is bad, it will **block developer to do the commit**, therefore developer can know the styling defect as soon as they attempt to commit and recommit again, faster than commit->push->travis->notice failure->update->recommit->repush

amazingandyyy: explanation as above

.eslintignore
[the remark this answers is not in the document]
amazingandyyy: updated.

amazingandyyy: updated.

package.json
dwalintukan: cool that makes sense to me. but, can't we just tell all our devs to use `lint:fix` before pushing or committing? it just seems redundant to have 3 of the same commands.

package.json
```diff
@@ -125,10 +127,19 @@
     "start:prod": "REACT_APP_ENV=prod node scripts/start.js",
     "build": "node scripts/build.js",
     "test": "node scripts/test.js --env=jsdom",
-    "lint:fix": "eslint --fix --",
+    "lint": "eslint .",
```
dwalintukan: do we need this? dont think we ever just run it without the --fix flag?

package.json
```diff
@@ -125,10 +127,19 @@
...
+    "lint:fix": "eslint . --fix",
     "extract-messages": "NODE_ENV=production babel ./src  --out-file /dev/null",
     "manage:translations": "NODE_ENV=production ./node_modules/.bin/babel-node scripts/translator.js",
-    "build:langs": "npm run extract-messages && npm run manage:translations"
+    "build:langs": "npm run extract-messages && npm run manage:translations",
+    "precommit": "lint-staged",
+    "prepush": "npm run lint:fix"
+  },
+  "lint-staged": {
+    "*.js": [
```
dwalintukan: does this follow the same rules as `.eslintignore`?

package.json
amazingandyyy: I add this because I think sometime engineer want to learn what they can do it better, otherwise the --fix always fix it for them. I myself used it for learning. but we can also remove it here if you prefer and expect engineer will use `eslint .` to debug.

package.json
amazingandyyy: it will not, but the `npm run lint:fix` will

PR #623, opened 2018-06-12
618 mobx 🔥

alex-cory:
This is just the base, should be ready to rock 💥

### DIFF
--- a/src/scenes/App/router.js
+++ b/src/scenes/App/router.js
@@ -6,7 +6,7 @@ import asyncComponent from '../../helpers/AsyncFunc';
 import { RouterPath } from '../../constants';
 
 
-const AppRouter = ({ url, langHandler, lang }) => {
+const AppRouter = ({ url }) => {
   // Remove trailing '/' from url so that we can use `${url}/topic` below
   if (url[url.length - 1] === '/') {
     url = url.slice(0, url.length - 1); // eslint-disable-line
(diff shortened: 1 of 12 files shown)


src/scenes/App/router.js
```diff
@@ -17,56 +17,54 @@ const AppRouter = ({ url, langHandler, lang }) => {
       <Route
         exact
         path={`${url}${RouterPath.qtumPrediction}`}
-        component={asyncComponent(() => import('../Dashboard/index'), langHandler, lang)}
+        component={asyncComponent(() => import('../Dashboard/index'))}
```
dwalintukan: much cleaner 👍

src/stores/UiStore.js
```diff
@@ -0,0 +1,38 @@
+import { observable, action, reaction, computed } from 'mobx';
+import moment from 'moment';
+import locales from '../languageProvider';
+
+
+export default class UIStore {
```
dwalintukan: nit: what do you think about calling it `LangStore`? or too specific?

src/stores/AppStore.js
```diff
@@ -0,0 +1,18 @@
+import { observable } from 'mobx';
+import UiStore from './UiStore';
+
+
+class AppStore {
+  @observable loading = true
```
dwalintukan: remove?

src/stores/UiStore.js
dwalintukan: if wont change, can we rename to `UiStore`? read somewhere that its normal to use camelcase even for abbreviated words.

alex-cory: Gah, I totally thought I made it `UiStore` lol. I wrote it `UIStore` at first but I thought I switched it 🤦‍♂️. So the reason I chose `UiStore` was from [this part of the docs](https://mobx.js.org/best/store.html). We can do `LangStore` but yeah I think it might be too specific

src/stores/AppStore.js
alex-cory: consider it removed!

src/stores/UiStore.js
alex-cory: Oh, I just realized I did `UiStore` everywhere except the definition name. Good catch

Issue #615, opened 2018-06-11
Socket hangup


dwalintukan:
Seems to be a socket issue due to cipher:
https://github.com/nodejs/node/issues/9845
https://stackoverflow.com/questions/42545683/socket-hang-up-at-tlssocket-onhangup-when-connecting-to-iis6-server-with-tls-1-0?rq=1

![image](https://user-images.githubusercontent.com/4350404/41225110-45fdb378-6d98-11e8-972b-9e78000eb2ff.png)

dwalintukan:
#628 might fix it.

dwalintukan:
verified fixed

PR #628, opened 2018-06-13
615 socket hangup

dwalintukan:
reference:
#615 
https://github.com/nodejs/node/issues/9845#issuecomment-264032107

### DIFF
--- a/src/network/httpRequest.js
+++ b/src/network/httpRequest.js
@@ -1,4 +1,18 @@
-import fetch from 'node-fetch';
+import axios from 'axios';
+import http from 'http';
+import https from 'https';
+
+let instance;
+
+export default function getAxios() {
+  if (!instance) {
+    instance = axios.create({
+      httpAgent: new http.Agent(),
+      httpsAgent: new https.Agent({ ciphers: 'DES-CBC3-SHA' }),
+    });
+  }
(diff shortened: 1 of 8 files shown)


src/network/httpRequest.js
```diff
@@ -1,33 +1,15 @@
...
- * Parses the HTTP response returned by a network request.
- * @param response {Object} A response from a network request.
- * @return {Object} The parsed JSON from the request.
- */
-function parseResponse(response) {
-  return response.json()
-    .then((res) => res.result || res)
-    .catch((err) => {
-      throw new Error(err);
+export default function getAxios() {
```
alex-cory: Can we export this as an object instead of a function so we can use it like
```js
import axios from 'httpRequest';
axios.get(Routes.whatever)
```
also, we might be able to do
```js
import { get } from 'httpRequest';
get(Routes.whatever)
```
Not sure on that last one 100% though^

Also, as far as I know, I don't think the `instance` variable is necessary. Unless you're globally attaching `instance`, each time you import `getAxios`, `instance === undefined` always.
```js
export default axios.create({
  ...
});
```

dwalintukan: fixed to export axios object.

PR #639, opened 2018-06-14
Add more folders to eslintignore

dwalintukan:


amazingandyyy:
good catch, also surprised that `--fix` silently fix everything in the build folder and not let us notice
.

PR #644, opened 2018-06-14
Change passphrase

henryzbw:
image
image
image
image


henryzbw:
#568

src/languageProvider/locales/en_US.json
```diff
@@ -25,6 +25,8 @@
   "cardInfo.topic": "Topic Created",
   "cardInfo.vote": "Voting",
   "cardInfo.withdraw": "Withdraw",
+  "changePassphrase.fail": "Passphrase Change Failed",
+  "changePassphrase.success": "Passphrase Change Successful.",
```
dwalintukan: nit: remove `.`

src/languageProvider/locales/en_US.json
```diff
@@ -157,10 +160,12 @@
   "str.minutes": "Minutes",
   "str.name": "Name",
   "str.newBlock": "New block comes, you can refresh to get new data",
+  "str.newPassphrase": "new passphrase",
```
dwalintukan: nit: capitalize

src/languageProvider/locales/en_US.json
```diff
@@ -157,10 +160,12 @@
...
   "str.notEnoughBot": "You don't have enough BOT",
   "str.notEnoughQtum": "You don't have enough QTUM",
   "str.notEnoughQtumAndBot": "You don't have enough QTUM or BOT",
-  "str.ok": "OK",
+  "str.ok": "Ok",
+  "str.oldPassphrase": "old passphrase",
```
dwalintukan: nit: capitalize

src/redux/App/reducer.js
```diff
@@ -13,6 +13,7 @@ const initState = new Map({
   appLocation: AppLocation.qtumPrediction,
   walletAddresses: [],
   lastUsedAddress: '',
+  changePassphraseResult: undefined,
```
dwalintukan: should be unnecessary if using `undefined` as default.

src/scenes/Wallet/components/ActionButtonHeader/index.js
```diff
@@ -75,6 +78,7 @@ export default class ActionButtonHeader extends Component {
         <EncryptDialog
           dialogVisible={encryptDialogVisible}
           onClose={() => { this.setState({ encryptDialogVisible: false }); }}
+          openPassphraseChangeDialog={() => { this.setState({ passphraseChangeDialogVisible: true }); }}
```
dwalintukan: can remove `{ }` and `;` for shortform

src/scenes/Wallet/components/ActionButtonHeader/index.js
```diff
@@ -84,6 +88,12 @@ export default class ActionButtonHeader extends Component {
           dialogVisible={restoreDialogVisible}
           onClose={() => { this.setState({ restoreDialogVisible: false }); }}
         />
+        <ChangePassphraseDialog
+          dialogVisible={passphraseChangeDialogVisible}
+          onClose={() => { this.setState({ passphraseChangeDialogVisible: false }); }}
+          closeEncryptDialog={() => { this.setState({ encryptDialogVisible: false }); }}
```
dwalintukan: shortform

src/scenes/Wallet/components/ChangPassphraseDialog/index.js
```diff
@@ -0,0 +1,100 @@
...
+  state = {
+    oldPassphrase: '',
+    newPassphrase: '',
+  };
+
+  submitPasswordChange = () => {
+    const { oldPassphrase, newPassphrase } = this.state;
+    this.props.changePassphrase(oldPassphrase, newPassphrase);
+    this.props.closeEncryptDialog();
+    this.props.onClose();
```
dwalintukan: destructure these: `const { changePassphrase, closeEncryptDialog, onClose } = this.props;`

src/scenes/Wallet/components/ChangPassphraseDialog/index.js
```diff
@@ -0,0 +1,100 @@
...
+          <div>
+            <Typography variant="title" ></Typography>
+            <TextField
+              autoFocus
+              margin="normal"
+              id="oldPassphrase"
+              label={intl.formatMessage(messages.oldPassphrase)}
+              type="password"
+              fullWidth
+              onChange={(e) => { this.setState({ oldPassphrase: e.target.value }); }}
```
dwalintukan: shortform

src/scenes/Wallet/components/ChangPassphraseDialog/index.js
```diff
@@ -0,0 +1,100 @@
...
+              error={_.isEmpty(oldPassphrase)}
+              required
+            />
+            <TextField
+              margin="normal"
+              id="newPassphrase"
+              label={intl.formatMessage(messages.newPassphrase)}
+              type="password"
+              fullWidth
+              onChange={(e) => { this.setState({ newPassphrase: e.target.value }); }}
```
dwalintukan: shortform

src/scenes/Wallet/components/ChangPassphraseDialog/index.js
```diff
@@ -0,0 +1,100 @@
...
+    id: 'str.newPassphrase',
+    defaultMessage: 'new passphrase',
+  },
+});
+
+
+@injectIntl
+@withStyles(null, { withTheme: true })
+@connect((state) => ({
+  walletAddresses: state.App.get('walletAddresses'),
```
dwalintukan: looks like this isnt used.

src/redux/App/reducer.js
alex-cory: yes, but it shows what's in the state. Also, probably better to use an empty string since that's what the type is right?

alex-cory: Just like in components we would do
```
state = {
  changePassphraseResult: '',
}
```
if we don't do that in components it will error. Redux doesn't require it, but it's nice to follow the same pattern so we can see what's in the state.

dwalintukan: if thats the case, i suggest we use the pattern of `undefined` meaning not used, and `null` as not assigned and default value.

henryzbw: the `changePassphraseResult` is an object  when we change passphrase.

PR #645, opened 2018-06-15
fix 书籍货币 to 数字货币

amazingandyyy:


amazingandyyy:
will land as soon as the travis passed

PR #646, opened 2018-06-15
Commit package-lock after npm install

dwalintukan:


amazingandyyy:
my conflicts with #645

amazingandyyy:
included it in #645

PR #700, opened 2018-06-19
Add missing phase

dwalintukan:


src/constants.js
```diff
@@ -26,11 +26,12 @@ module.exports = {
   },
 
   Phases: {
-    bet: 'bet',
-    vote: 'vote',
-    setResult: 'setResult',
-    finalize: 'finalize',
-    withdraw: 'withdraw',
+    BETTING: 'betting',
```
alex-cory: can we make these lowercase? Sometimes I feel like we're switching between uppercase, lowercase, pascal case, and camelCase and it's nice to just have the keys of the maps be the same.
For instance:
if I want to do this
```js
const phase = getPhase(oracle);
const appLocation = AppLocation[phase]
```
Or vice versa. I can't really do either because all our keys are different amongst our constants. It would be great if we could make the keys the same so we could reuse them in various parts of the app yeah?
```js
Phases: {
  betting: 'betting'
  ...
},
AppLocation: {
  betting: '/bet'
  ...
}
```
Also, I had `betting` etc, but switched to `bet` for brevity. So chill with either on that front :)

alex-cory: Does what I'm suggesting make sense? I can provide other examples if it doesn't :)

dwalintukan: how do you feel about ALL UPPERCASE for these constants? that's generally a recognized standard cross-language convention.

amazingandyyy: I personally use UPPERCASE for this kind of constants as well.  but if so, shall we refactor whole `constants.js` and all related files in this PR?

dwalintukan: I have updated the phases to all uppercase.

@amazingandyyy I will make a separate PR for refactoring the rest of this. When doing PRs, let's try to keep the scope of it as on point as possible.

dwalintukan:
sounds good

PR #701, opened 2018-06-19
Change port number to 8989

dwalintukan:
Depends on: https://github.com/bodhiproject/bodhi-server/pull/40/files


amazingandyyy:
need to ship at the same time as https://github.com/bodhiproject/bodhi-server/pull/40

PR #714, opened 2018-06-21
Mobx to event cards

amazingandyyy:
finish refactor the first page of the dashboard and rename it `QtumPrediction`


src/stores/QtumPredictionStore.js
```diff
@@ -0,0 +1,72 @@
+import { observable, action, runInAction, reaction } from 'mobx';
+import _ from 'lodash';
+import { Token, OracleStatus, AppLocation } from '../constants';
+import { queryAllOracles } from '../network/graphQuery';
+import Oracle from './models/Oracle';
+
+
+export default class AllPredictionsStore {
```
dwalintukan: QtumPredictionStore?

src/stores/QtumPredictionStore.js
```diff
@@ -0,0 +1,72 @@
...
+  @observable hasMore = true
+  @observable skip = 0
+  limit = 50
+
+  constructor(app) {
+    this.app = app;
+    reaction(
+      () => this.app.sortBy, // when 'sortBy' changes
+      () => {
+        // and we're on the AllEvents page
```
dwalintukan: QtumPrediction

src/stores/QtumPredictionStore.js
```diff
@@ -0,0 +1,72 @@
...
+  @observable loadingMore = false
+  @observable list = []
+  @observable hasMore = true
+  @observable skip = 0
+  limit = 50
+
+  constructor(app) {
+    this.app = app;
+    reaction(
+      () => this.app.sortBy, // when 'sortBy' changes
```
dwalintukan: suggestion: why don't we have separate `sortBy` per each page? that way if the user navigates out and back in to the page it will stay on whichever one they chose previously.

src/stores/QtumPredictionStore.js
```diff
@@ -0,0 +1,72 @@
...
+  @action.bound
+  async init(limit = this.limit) {
+    if (limit === this.limit) {
+      this.skip = 0;
+    }
+    this.hasMore = true;
+    this.app.ui.location = AppLocation.qtumPrediction;
+    this.list = await this.fetchQtumPredictions(limit);
+    runInAction(() => {
+      this.skip += this.limit;
```
dwalintukan: shouldn't it use the local var `limit` instead of `this.limit` since if its passed in as a param it might be different?

src/stores/QtumPredictionStore.js
amazingandyyy: updated. thank you

dwalintukan:
LGTM after the changes ^. pre-approving now.

src/stores/QtumPredictionStore.js
amazingandyyy: move this to #718, want to land this first.

PR #721, opened 2018-06-22
refactor encryptDialog to mobx

henryzbw:


src/scenes/Wallet/components/EncryptDialog/index.js
```diff
@@ -71,7 +58,7 @@ export default class EncryptDialog extends Component {
                 type="password"
                 fullWidth
                 error={_.isEmpty(passphrase)}
-                onChange={(e) => this.setState({ passphrase: e.target.value })}
+                onChange={(e) => onPassphraseChange(e.target.value)}
```
amazingandyyy: @alex-cory :
```js
const { wallet, wallet: {walletEncrypted} } = this.props.store;
onChange={(e) => wallet.passphrase = e.target.value}`;
```
change to this.

src/stores/WalletStore.js
```diff
@@ -1,6 +1,33 @@
...
+      const errorObject = {
+        route: Routes.api.encryptWallet,
+        message: error.message,
+      };
+      this.error = errorObject;
+    }
+  }
+
+  @action
+  onPassphraseChange = (passphrase) => {
```
amazingandyyy: not really necessary here. 🥇

src/scenes/Wallet/components/EncryptDialog/index.js
henryzbw:
```js

henryzbw: better keep all the store change function inside the store file. Otherwise may cause confusion.

PR #724, opened 2018-06-24
675 mobx wallet history

dwalintukan:


src/stores/WalletHistoryStore.js
```diff
@@ -0,0 +1,63 @@
...
+  @observable list = []
+  @observable orderBy = 'createdTime'
+  @observable direction = SortBy.Descending
+  @observable limit = 500
+  @observable skip = 0
+  @observable perPage = 10
+  @observable page = 0
+  @observable expanded = []
+
+  @action.bound
```
henryzbw: @action 
```
Note: don't use action.bound with arrow functions; arrow functions are already bound and cannot be rebound.
```
https://mobx.js.org/refguide/action.html

src/stores/WalletHistoryStore.js
```diff
@@ -0,0 +1,63 @@
...
+      this.fullList = _.map(result, (tx) => new Transaction(tx));
+      this.list = _.slice(this.fullList, 0, this.perPage);
+    } catch (error) {
+      console.error(error); // eslint-disable-line
+      this.fullList = [];
+      this.list = [];
+    }
+  }
+
+  @action.bound
```
henryzbw: same as above

src/stores/WalletHistoryStore.js
```diff
@@ -0,0 +1,63 @@
...
+  onPageChange = (page) => {
+    this.expanded = [];
+    this.page = page;
+
+    const start = this.page * this.perPage;
+    this.list = _.slice(this.fullList, start, start + this.perPage);
+  }
+
+  @action.bound
```
henryzbw: same as above

src/stores/WalletHistoryStore.js
```diff
@@ -0,0 +1,63 @@
...
+  onPerPageChange = (perPage) => {
+    this.expanded = [];
+    this.perPage = perPage;
+    this.list = _.slice(this.fullList, 0, this.perPage);
+  }
+
+  @action.bound
```
henryzbw: same as above

src/stores/WalletHistoryStore.js
```diff
@@ -0,0 +1,63 @@
...
+  onExpandedChange = (expanded) => {
+    this.expanded = expanded;
+  }
+
+  @action.bound
```
henryzbw: same as above

src/stores/WalletHistoryStore.js
```diff
@@ -0,0 +1,63 @@
...
+      this.list = _.slice(this.fullList, 0, this.perPage);
+    } catch (error) {
+      console.error(error); // eslint-disable-line
+      this.fullList = [];
+      this.list = [];
+    }
+  }
+
+  @action.bound
+  onPageChange = (page) => {
```
alex-cory: Ok, so just a thought here on how I would write this. I kinda don't like having to manage `state` of items in an array like this. I would do something like

`/WalletHistory/HistoryItem/index.js` or something along those lines
and do something like
```js
@observer
class HistoryItem extends Component {
  @observable open = false // OR state = { expanded: false }
  toggle = () => this.open = !this.open
  render() {
    return (
      <Fragment>
        <Item onClick={this.toggle}>
        </Item>
        {this.open && (
          <Drawer></Drawer>
        )}
      </Fragment>
    )
  }
}
```

src/stores/WalletHistoryStore.js
alex-cory: same as `@action` comment above

src/stores/WalletHistoryStore.js
alex-cory: same as `@action` comment above

src/stores/WalletHistoryStore.js
alex-cory: same as `@action` comment above

src/stores/WalletHistoryStore.js
alex-cory: same as `@action` comment above

src/stores/WalletHistoryStore.js
alex-cory: Then we can remove all the `this.expanded = []` from all these methods

src/stores/WalletHistoryStore.js
```diff
@@ -0,0 +1,63 @@
...
+    this.expanded = [];
+    this.page = page;
+
+    const start = this.page * this.perPage;
+    this.list = _.slice(this.fullList, start, start + this.perPage);
```
alex-cory: This is a great place for a `reaction`
WalletHistory/index.js
```js
          <TableFooter
              fullList={walletHistory.fullList}
              perPage={walletHistory.perPage}
              page={walletHistory.page}
              onPageChange={(page) => walletHistory.page = page}
              onPerPageChange={(perPage) => walletHistory.perPage = perPage}
            />
```
HistoryStore.js
```js
class HistoryStore {
  @observable fullList
  @observable page
  @observable perPage
  @observable list = []
  constructor() {
    reaction(
      () => this.perPage, // when perPage changes
      () => this.list = _.slice(this.fullList, 0, this.perPage);
    )
    reaction(
      () => this.page, // when page changes
      () => {
        const start = this.page * this.perPage;
        this.list = _.slice(this.fullList, start, start + this.perPage)
      }
    ) 
  }
}
```

src/stores/WalletHistoryStore.js
```diff
@@ -0,0 +1,63 @@
...
+
+  @action.bound
+  onPerPageChange = (perPage) => {
+    this.expanded = [];
+    this.perPage = perPage;
+    this.list = _.slice(this.fullList, 0, this.perPage);
+  }
+
+  @action.bound
+  onExpandedChange = (expanded) => {
```
alex-cory: can remove if we make the `HistoryItem/index.js` component :)

src/stores/WalletHistoryStore.js
```diff
@@ -0,0 +1,63 @@
...
+    this.expanded = expanded;
+  }
+
+  @action.bound
+  onSortingChange = (orderBy, direction) => {
```
alex-cory: this is a great place to use a `reaction`, but this is also fine :)

src/stores/models/Transaction.js
```diff
@@ -0,0 +1,43 @@
...
+
+  constructor(transaction, app) {
+    Object.assign(this, transaction);
+    this.app = app;
+    this.gasLimit = Number(this.gasLimit);
+    this.gasPrice = Number(this.gasPrice);
+    this.fee = gasToQtum(this.gasUsed);
+
+    if (this.token === Token.Bot) {
+      if (this.type !== TransactionType.ApproveCreateEvent
```
alex-cory: nit: `![ApproveCreateEvent, ApproveSetResult, ApproveVote].some(this.type)` basically the logic is "if not 1 of these is true"

src/stores/models/Transaction.js
```diff
@@ -0,0 +1,43 @@
+import { Token, TransactionType } from '../../constants';
+import { gasToQtum, satoshiToDecimal } from '../../helpers/utility';
+
+export default class Transaction {
```
alex-cory: 🔥

src/stores/WalletHistoryStore.js
```diff
@@ -0,0 +1,63 @@
...
+  @observable expanded = []
+
+  @action.bound
+  queryTransactions = async (orderBy = this.orderBy, direction = this.direction, limit = this.limit, skip = this.skip) => {
+    try {
+      const filters = [{ type: TransactionType.Transfer }];
+      const orderByObj = { field: orderBy, direction };
+
+      const result = await queryAllTransactions(filters, orderByObj, limit, skip);
+      this.fullList = _.map(result, (tx) => new Transaction(tx));
```
alex-cory: we should have a `runInAction` wrapped around the 2 lines after line 26 since we're modifying an `@observable` inside an `async @action`

alex-cory: basically we always want to do this after we `await` in an `async @action`

alex-cory: I know... kinda weird, but this will keep from weird bugs happening (like the list and the fullList might not update)

alex-cory: Also, typically I like to split my `api request` methods and my `setter methods`.
```js
@action.bound
async init () {
  ...
  const transactions = await this.fetchTransactions()
  runInaction(() => {
    this.fullList = transactions
    this.list = ...
  })
}

@action.bound
async loadMore() {
  const transactions = await this.fetchTransactions(this.skip, this.limit)
  runInaction(() => {
    this.fullList = transactions
    this.list = ...
  })
}

fetchTransactions(skip, limit) {
  return queryAllTransactions(...)
}
```
because sometimes we might need to make an `api call` in another method in our store

src/stores/WalletHistoryStore.js
dwalintukan: oops! i can't read italic text. fixed

src/stores/WalletHistoryStore.js
dwalintukan: fixed

src/stores/WalletHistoryStore.js
dwalintukan: fixed

src/stores/WalletHistoryStore.js
dwalintukan: fixed

src/stores/WalletHistoryStore.js
dwalintukan: fixed

src/scenes/Wallet/components/History/index.js
```diff
@@ -11,96 +12,79 @@ import {
...
-        transactions: sorted,
-      });
-    }
+  componentDidMount() {
+    this.props.store.walletHistory.queryTransactions();
   }
 
-  componentDidUpdate(prevProps, prevState) {
-    const { skip } = this.state;
+  componentWillReceiveProps(nextProps) {
```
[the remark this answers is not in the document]
dwalintukan: is it cool if we just leave this for now? Henry is doing the `txReturn` -> MobX change which directly affects this. Also once we put `syncBlockNum` in MobX this will change as well. We can change this into a reaction in the store instead.

src/stores/WalletHistoryStore.js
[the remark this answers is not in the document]
dwalintukan: fixed

src/stores/WalletHistoryStore.js
dwalintukan: fixed

src/stores/WalletHistoryStore.js
dwalintukan: fixed

src/stores/WalletHistoryStore.js
dwalintukan: fixed

src/stores/WalletHistoryStore.js
dwalintukan: fixed

src/stores/WalletHistoryStore.js
dwalintukan: done, makes sense. I am keeping the expanded in the component state. is that ok?

src/stores/WalletHistoryStore.js
dwalintukan: good call 👍

src/stores/WalletHistoryStore.js
dwalintukan: done

src/stores/WalletHistoryStore.js
dwalintukan: done

src/stores/models/Transaction.js
dwalintukan: decided to use `_.includes`

src/stores/models/Transaction.js
dwalintukan: 🐶

src/stores/WalletHistoryStore.js
dwalintukan: added runInActions. for splitting the api request, I feel it makes sense to just have the `queryTransactions` call set the list since this page doesn't have pagination and it's only called once.

src/scenes/Wallet/components/History/index.js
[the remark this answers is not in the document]
alex-cory: yeah that's fine! I will say the implementation is just
```js
componentDidUpdate(prevProps) {
    const { txReturn, syncBlockNum, store: { walletHistory: { queryTransactions } } } = prevProps;

    if ((txReturn && !this.props.txReturn) || (syncBlockNum !== this.props.syncBlockNum)) {
      queryTransactions();
    }
}
```

src/stores/models/Transaction.js
```diff
@@ -0,0 +1,44 @@
...
+  oracleAddress = ''
+  name = ''
+  optionIdx = 0
+  token = ''
+  amount = ''
+  topic
+
+  constructor(transaction, app) {
+    Object.assign(this, transaction);
+    this.app = app;
```
alex-cory: so we're not actually passing the `app` store down. We should either pass it in, or remove this prop for now until we need it :)

src/stores/WalletHistoryStore.js
```diff
@@ -0,0 +1,108 @@
...
+
+      runInAction(() => {
+        this.fullList = [];
+        this.list = [];
+      });
+    }
+  }
+
+  @action
+  onPageChange = (page) => {
```
alex-cory: so, I would prefer we not have `setter` methods like this, but if we're hard on it then it's not the end of the world. I think it just removes redundant code and increases readability of the tests. For instance
```js
describe('test something', async () => {
  const appStore = new AppStore()
  const { walletHistory } = appStore
  await walletHistory.init()
  walletHistory.page = 1 // <- setting this way I think is more clear/easier to read than
  walletHistory.onPageChange(1) // as opposed to this
  expect(walletHistory.list).toEqual(...)
})

```

src/scenes/Wallet/components/History/components/HistoryItem/index.js
```diff
@@ -0,0 +1,66 @@
+import React, { Component, Fragment } from 'react';
+import { TableCell, TableRow, withStyles } from '@material-ui/core';
+import cx from 'classnames';
+
+import styles from './styles';
+import { getShortLocalDateTimeString } from '../../../../../../helpers/utility';
+import TransactionHistoryID from '../../../../../../components/TransactionHistoryAddressAndID/id';
+import TransactionHistoryAddress from '../../../../../../components/TransactionHistoryAddressAndID/address';
+
+export default class HistoryItem extends Component {
```
alex-cory: 🔥

src/scenes/Wallet/components/History/index.js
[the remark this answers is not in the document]
dwalintukan: oh ok ill change it if its that easy of a change

src/stores/models/Transaction.js
dwalintukan: fixed, accidentally left that in there :P

src/stores/WalletHistoryStore.js
dwalintukan: oh i see. i'm ok with not having setter methods too. didn't know you wanted to just set everything directly from the comp. no problem with me, fixed

src/stores/models/Transaction.js
amazingandyyy: 👹

PR #725, opened 2018-06-25
mobx actionButtonHeader

henryzbw:


src/scenes/Wallet/components/ActionButtonHeader/index.js
```diff
@@ -1,15 +1,13 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { connect } from 'react-redux';
 import { injectIntl, defineMessages } from 'react-intl';
 import { withStyles, Button, Tooltip } from '@material-ui/core';
-
+import { inject, observer } from 'mobx-react';
```
dwalintukan: put this in the top imports. the order goes: external imports -> project imports

src/scenes/Wallet/components/ActionButtonHeader/index.js
```diff
@@ -40,25 +38,13 @@ const messages = defineMessages({
   },
 });
 
+@inject('store')
+@observer
```
dwalintukan: move `inject('store') and observer` to the very last injections. this can sometimes cause really weird bugs if its above others.

src/stores/WalletStore.js
```diff
@@ -35,4 +35,17 @@ export default class WalletStore {
...
+  @action
+  backupWallet = async () => {
+    try {
+      await axios.post(Routes.api.backupWallet);
+    } catch (error) {
+      const errorObject = {
+        route: Routes.api.backupWallet,
+        message: error.message,
+      };
+      this.error = errorObject;
```
dwalintukan: FYI, looks like showing this error obj in the `ErrorDialog` has a dependency on doing the MobX refactor for the `ErrorDialog`. the errors are being stored here, but aren't being used. So if backup or encrypt fail it is not being caught.

src/scenes/Wallet/components/ActionButtonHeader/index.js
henryzbw: @dwalintukan oh, i c. seems like the linter rule for this gets disabled?

dwalintukan: @henryzbw technically its fine. just a nitpick that the external imports was grouped with the project imports.

dwalintukan:
you can use the `ErrorDialog` now after rebase. i've already done it for the encryptWallet function.

src/stores/WalletStore.js
henryzbw: @dwalintukan just check the code, this encrypt doesn't use the `ErrorDialog`, i used `EncryptStatusDialog` to display the success or failed by check the `encryptResult `object

dwalintukan: @henryzbw hm ok. can you use the `ErrorDialog` if it errors and `EncryptStatusDialog` for success? I want to formalize all errors that need a dialog to use the `ErrorDialog`.

src/scenes/Wallet/components/ActionButtonHeader/index.js
```diff
@@ -1,15 +1,14 @@
...
+import { inject, observer } from 'mobx-react';
 import { withStyles, Button, Tooltip } from '@material-ui/core';
-
 import EncryptDialog from '../EncryptDialog';
 import EncryptStatusDialog from '../EncryptStatusDialog';
 import RestoreWalletDialog from '../RestoreWalletDialog';
 import ChangePassphraseDialog from '../ChangPassphraseDialog';
 import ChangePassphraseStatusDialog from '../ChangePassphraseStatusDialog';
-import appActions from '../../../../redux/App/actions';
+import ErrorDialog from '../../../../components/ErrorDialog';
```
dwalintukan: This should not be necessary. The `ErrorDialog` is already in `App/index.js` which is a top level component that should handle this. All you have to do is set the `error` in the mobx store and it should show up. This is duplicating the dialog.

src/scenes/Wallet/components/ActionButtonHeader/index.js
```diff
@@ -68,26 +55,25 @@ export default class ActionButtonHeader extends Component {
...
-        <EncryptStatusDialog
-          onClose={() => { this.setState({ encryptDialogVisible: false }); }}
-          encryptResult={encryptResult}
-        />
+        <EncryptStatusDialog />
         <RestoreWalletDialog
           dialogVisible={restoreDialogVisible}
           onClose={() => { this.setState({ restoreDialogVisible: false }); }}
         />
+        <ErrorDialog />
```
dwalintukan: remove

Issue #726, opened 2018-06-26
Crushed at the initial run when using an OS with a foreign locale setting


ippudo:
It happened when I tried to run Bodhi-ui in an operating system with a locale setting other than Korean/Chinese/English at the first time.

PR #729, opened 2018-06-26
#726 Fixed Null locale init for non ZH/KO/EN users

ippudo:
OS with language other than ZH/KO/EN will not return EN-US as well.

Please don't hesitate to tell me if I did something wrong.


amazingandyyy:
fix #726

Issue #737, opened 2018-07-01
Entering the amount is broken and doesnt accept the input


dwalintukan:
Believe these are the steps to repro:
1. Go into an event in Qtum Prediction
2. Make a bet
3. wait for success
4. try to do another bet
5. unable to press the bet button

dwalintukan:
cant seem to repro this.

PR #738, opened 2018-07-01
Hot fixes

alex-cory:
Once we get this mobx refactor done, we shouldn't have bugs like this (bet amount input not working) occur anymore since the logic for voting will be in a separate page.


alex-cory:
https://github.com/bodhiproject/bodhi-ui/issues/737

src/constants.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
     vote: 'Vote',
     wallet: 'Wallet',
     myActivities: 'MyActivities',
-    resultSetting: 'ResultSetting',
+    resultSet: 'ResultSetting',
```
alex-cory: I think eventually we should change this to reflect it's value like the rest of these

dwalintukan: I have an issue open to refactor all these. we should just use ALL_CAPS for key/value.

src/scenes/Event/components/EventOption/index.js
```diff
@@ -53,10 +54,12 @@ export default class EventOption extends Component {
 
 
   handleAmountBlur = ({ target: { value } }) => {
-    let { amount, consensusThreshold, onAmountChange } = this.props; // eslint-disable-line
-    [amount, consensusThreshold] = [parseInt(amount, 10), parseInt(consensusThreshold, 10)];
-    if (amount + value > consensusThreshold) {
-      onAmountChange(consensusThreshold - amount);
+    let { phase, amount, consensusThreshold, onAmountChange } = this.props; // eslint-disable-line
+    if (phase === Phases.VOTING) {
+      [amount, consensusThreshold] = [parseInt(amount, 10), parseInt(consensusThreshold, 10)];
```
dwalintukan: we should be parsing these to floats. the core issue that was the problem before was this situation:

- say the threshold is 100 BOT
- someone previously voted 99.99999999
- you want to vote to the threshold, but you would have to enter 0.00000001 in the amount
- the fix should be that you can enter 1 (or anything higher) and it would auto adjust to the amount (0.00000001) and let you know with some message saying `You can only vote up to the threshold, etc`

alex-cory: Yeah that's what I thought we discussed and then you liked my onBlur idea where it would automatically correct it to the right amount. Right?

alex-cory: I see the problem though. Fixing. Apologies

alex-cory: this is now fixed

PR #739, opened 2018-07-02
walletUnlockedUntil & WalletUnlockDialog mobx

dwalintukan:


src/scenes/App/globalHub.js
```diff
@@ -116,11 +103,6 @@ export default class GlobalHub extends Component {
       || (syncPercent === 100 && syncBlockNum !== nextProps.syncBlockNum)) {
       getPendingTransactions();
     }
-
-    // Show the GlobalSnackbar with wallet unlocked message
-    if (walletUnlockedUntil === 0 && nextProps.walletUnlockedUntil > 0) {
```
henryzbw: so cool. I was blocked by this. Didn't realize we can just create an store for the walletUnlockDialog and then set the dialog open props by using `store.isVisible`.

Issue #713, opened 2018-06-21
Add number of Connected Peers to Loading Screen and Bottom Bar


dwalintukan:
Need to show the number of connected peers at all times like QT Wallet.

PR #741, opened 2018-07-02
Global store mobx

dwalintukan:
- Renamed `PubsubStore` to `GlobalStore` for any globally used values
- This is for setting up `syncInfo` calls using MobX and storing it. Once the refactors for other components where they need the `syncInfo attributes: syncPercent, syncBlockNum, or syncBlockTime` you can use this `GlobalStore`'s values.

### DIFF
--- a/src/components/BottomBar/index.js
+++ b/src/components/BottomBar/index.js
@@ -1,7 +1,7 @@
 /* eslint-disable react/no-unused-state */
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { connect } from 'react-redux';
+import { inject, observer } from 'mobx-react';
 import { Paper, Grid, Typography, withStyles } from '@material-ui/core';
 import { CheckCircle as CheckCircleIcon, RemoveCircle as RemoveCircleIcon } from '@material-ui/icons';
 import { FormattedMessage, injectIntl } from 'react-intl';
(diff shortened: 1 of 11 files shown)


amazingandyyy:
this will also relate to my connection-count PR. #713

src/appProvider.js
```diff
@@ -20,7 +20,7 @@ export const AppProvider = observer(({ xstore, store }) => (
   <Provider store={store}>
     <MobxProvider store={xstore}>
       <MuiThemeProvider theme={bodhiTheme}>
-        <IntlProvider locale={xstore.ui.locale} messages={xstore.ui.localeMessages}>
+        <IntlProvider locale={xstore.global.locale} messages={xstore.global.localeMessages}>
```
alex-cory: can we please leave the `PubSubStore` and the `UiStore` separate?
[UiStore Example](https://mobx.js.org/best/store.html)
image

The PubSubStore is used to modify other stores when a new block comes in. For instance, let's say we're on the `QTUM Prediction` page, and we get some new data, we then call `this.app.qtumPrediction.init()` from the `PubSubStore`. These are just 2 very separate concerns.

alex-cory: We can also put `Error` in the `UiStore` since it's the `state of a global overlay`

alex-cory: If we begin to have more and more realtime things happen with our app, this `GlobalStore` is going to get pretty long. I can show you an example from my last company if you'd like

PR #754, opened 2018-07-06
Connection count

amazingandyyy:
add connection count to bottomBar


src/components/BottomBar/index.js
```diff
@@ -38,12 +33,11 @@ export default class BottomBar extends Component {
 
   render() {
     const { classes } = this.props;
-    const { syncBlockNum, syncBlockTime } = this.props.store.global;
-
+    const { syncBlockTime, syncBlockNum, peerNodeCount } = this.props.store.global || { syncBlockTime: 1, syncBlockNum: 1, peerNodeCount: 1 };
```
dwalintukan: lets not set the defaults here if they are undefined. the store defaults should be sufficient enough.

src/components/BottomBar/index.js
```diff
@@ -16,10 +15,6 @@ import { getShortLocalDateTimeString } from '../../helpers/utility';
 @inject('store')
 @observer
 export default class BottomBar extends Component {
-  static propTypes = {
-    classes: PropTypes.object.isRequired,
```
dwalintukan: this is required. the `Paper` in the render function uses `classes`.

src/components/BottomBar/index.js
amazingandyyy: good point

src/components/BottomBar/index.js
amazingandyyy: oh ok, I thought we are going to remove all propTypes, aren't we?

dwalintukan: @amazingandyyy classes is a proptype from the material UI lib. its not something that would go into the store. same goes for things like localization proptypes, etc.

PR #755, opened 2018-07-06
750 new block fix

dwalintukan:


dwalintukan:
@amazingandyyy The errors in the next function are free the graphql side. The server returns a 200 but with an errors obj. This is different than the error func of the subscribe method.

I spent a while debugging it. Couldn't find specific docs about it.

PR #759, opened 2018-07-09
Add connection to loading page and update the bottom one

amazingandyyy:


amazingandyyy:
#713

amazingandyyy:
screen shot 2018-07-09 at 16 14 26

amazingandyyy:
screen shot 2018-07-09 at 16 16 28
screen shot 2018-07-09 at 16 16 33

src/scenes/App/components/Loader/index.js
```diff
@@ -37,12 +36,9 @@ export default class Loader extends Component {
...
           </div>
           <div className={classes.loaderPercentWrapper}>
-            <Typography variant="display1" className={classes.loaderPercent}>{syncPercent}</Typography>%
-            <p>
-              <FormattedMessage id="str.blockSync" defaultMessage="Blockchain syncing..." />
-              &nbsp;
-              <FormattedMessage id="str.wait" defaultMessage="Please wait." />
-            </p>
+            <Typography variant="display1" className={classes.loaderPercent}>{syncPercent}</Typography>
+            <p>%</p>
```
henryzbw: keep `%` in the same line with `syncPercent`

amazingandyyy: why, <p></p> is a different component, and they will still be on the same line.

Issue #748, opened 2018-07-03
Oracle result setting shows upcoming label


henryzbw:
![image](https://user-images.githubusercontent.com/10339131/42248608-4c7dca9a-7edb-11e8-9a9e-5b8c27bf5730.png)

ippudo:
Fixed at #760

PR #760, opened 2018-07-09
748 isupcoming issue

ippudo:
#748 One-line fix to detect whether the result setter is user himself

### DIFF
--- a/src/stores/models/Oracle.js
+++ b/src/stores/models/Oracle.js
@@ -60,7 +60,7 @@ export default class Oracle {
     const isPending = oracle.transactions.some(({ type, status }) => pendingTypes.includes(type) && status === Pending);
     this.unconfirmed = (!oracle.topicAddress && !oracle.address) || isPending;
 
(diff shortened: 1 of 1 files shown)


src/stores/models/Oracle.js
```diff
@@ -60,7 +60,7 @@ export default class Oracle {
     const isPending = oracle.transactions.some(({ type, status }) => pendingTypes.includes(type) && status === Pending);
     this.unconfirmed = (!oracle.topicAddress && !oracle.address) || isPending;
 
-    this.isUpcoming = this.phase === RESULT_SETTING && oracle.status === OracleStatus.WaitResult;
+    this.isUpcoming = this.phase === RESULT_SETTING && oracle.status === OracleStatus.WaitResult && (app.wallet.addresses.filter((addr) => ('address' in addr) && (addr.address === this.resultSetterQAddress)).length === 0);
```
amazingandyyy: what is `'address' in addr` ?

amazingandyyy: will that return a truth value?

amazingandyyy: `addr['address']` ?

amazingandyyy:
which issue is this PR solving?

Issue #690, opened 2018-06-19
MobX: Components/SelectAddressDialog


dwalintukan:

ippudo:
Um I'm not sure but it seems like nothing to do with this component too. Do you have any suggestion about this component?

ippudo:
TODO: dialogVisible

dwalintukan:
@ippudo you should create a store for this component and it will have the `isVisible` in the store to control if it's showing or not. You will then observe this var in the component.

ippudo:
Complete at #763

PR #763, opened 2018-07-10
690 select address dialog

ippudo:
Mobx SelectAddressDialog(#690)


amazingandyyy:
will have huge conflicts with master, suggest to create a new branch and PR

PR #765, opened 2018-07-11
688 MobX navbar

### LINKED ISSUE #688: MobX: Components/NavBar

amazingandyyy:
fix #688. 2 less component is connecting to redux hoo


src/components/NavBar/store.js
```diff
@@ -1,20 +1,80 @@
...
   myActivitesCount: 0, // INIT_VALUESial loaded state
 };
 
 
-export default class NavBarStore {
-  @observable myActivitesCount = INIT_VALUES.myActivitesCount
+export default class {
+  @observable ResultSettingCount = INIT_VALUES.ResultSettingCount
+  @observable FinalizeCount = INIT_VALUES.FinalizeCount
+  @observable WithdrawCount = INIT_VALUES.WithdrawCount
```
dwalintukan: camelCase for all 3

src/components/NavBar/store.js
```diff
@@ -1,20 +1,80 @@
...
+        const { voterQAddress, topicAddress, optionIdx } = vote;
+        if (!_.find(accumulator, { voterQAddress, topicAddress, optionIdx })) accumulator.push(vote);
+        return accumulator;
+      }, []);
+
+      _.each(votes, ({ topicAddress, optionIdx }) => {
+        topicFilters.push({ status: OracleStatus.WITHDRAW, address: topicAddress, resultIdx: optionIdx });
+      });
+      let topics = await queryAllTopics(topicFilters);
+      topics = _.uniqBy(topics, 'txid').map((topic) => new Topic(topic, this.app));
```
dwalintukan: `uniqBy` should not be necessary here. There should not be any duplicate Topics in any case. If there is, we should actually let the error show up so we can know its happening. But it should never happen anyways.

Also, no need to map it to a `Topic` obj. We are only using the length of the array, not the actual object.

src/components/NavBar/store.js
```diff
@@ -1,20 +1,80 @@
...
+      const oracleSetFilters = [{ token: Token.QTUM, status: OracleStatus.OPEN_RESULT_SET }];
+      _.each(action.walletAddresses, (item) => {
+        oracleSetFilters.push({
+          token: Token.QTUM,
+          status: OracleStatus.WAIT_RESULT,
+          resultSetterQAddress: item.address,
+        });
+      });
+      let oracles = await queryAllOracles(oracleSetFilters);
+      oracles = _.uniqBy(oracles, 'txid').map((oracle) => new Oracle(oracle, this.app));
```
dwalintukan: Same comment as above. `uniqBy` not necessary. If there is duplicates, we should see the error so we can fix it instead of failing silently. This can lead to harder debugging later, or we would never know something is wrong with the BE.

No need to map to Oracle, we are only using the length.

src/components/NavBar/store.js
```diff
@@ -1,20 +1,80 @@
...
+      this.ResultSettingCount = oracles.length;
+
+      // Get finalize items
+      const oracleFinalizeFilters = [{ token: Token.BOT, status: OracleStatus.WAIT_RESULT }];
+      const oraclesForFinalize = await queryAllOracles(oracleFinalizeFilters);
+      oracles = _.uniqBy(oraclesForFinalize, 'txid').map((o) => new Oracle(o, this.app));
```
dwalintukan: remove line

src/components/NavBar/store.js
```diff
@@ -1,20 +1,80 @@
...
 
-  // @action.bound
-  // async getMyActivitiesCount() {
+  @action
+  init() {
+    this.getCount();
+  }
+
+  @action.bound
+  async getCount() {
```
dwalintukan: Can you actually move this to the `GlobalStore`? This call is in `GlobalHub` and the counts are not only used in the `NavBar`. It is also used in the `Activites` component in the tab counts. 

Once moved to GlobalStore, believe you can remove it entirely from the `GlobalHub`.

src/components/NavBar/store.js
dwalintukan: also, missing the `INIT_VALUES` for all 3

src/components/NavBar/store.js
```diff
@@ -1,20 +1,80 @@
...
-    // this.myActivitesCount = await this.getMyActivitiesCount();
+  constructor(app) {
+    this.app = app;
   }
 
-  // @action.bound
-  // async getMyActivitiesCount() {
+  @action
+  init() {
+    this.getCount();
```
dwalintukan: you are missing the `reactions` for calling `getCount()` again on `syncBlockNum` and `walletAddresses` change. see `globalHub.js L83`.

amazingandyyy: will updated

src/components/NavBar/store.js
amazingandyyy: hm, I simply follow the logic from the old saga, are you sure we don't need it? I join the team after, sometimes just cannot rethink everything, and if there are no unique thing, it's db problem, should not bother user, we should save it to backlog?

src/components/NavBar/store.js
amazingandyyy: if it's not necessary, why do we do it for rendering the cards and why we have them in the old saga?

src/components/NavBar/index.js
```diff
@@ -104,9 +99,9 @@ const BotCourt = observer(({ classes, store: { ui } }) => (
   </NavLink>
 ));
 
-const Wallet = ({ classes, walletAddresses }) => {
-  const totalQTUM = _.sumBy(walletAddresses, ({ qtum }) => qtum).toFixed(2) || '0.00';
-  const totalBOT = _.sumBy(walletAddresses, ({ bot }) => bot).toFixed(2) || '0.00';
+const Wallet = ({ classes, store: { wallet } }) => {
```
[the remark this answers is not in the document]
amazingandyyy: updated

src/components/NavBar/store.js
amazingandyyy: what a crap, sorry about this

src/components/NavBar/store.js
amazingandyyy: I know other places will use it, but I haven't decided where to put, I think putting it to GlobalStore is bad, I was thinking about putting it to a new store called ActivitiesStore(was thinking to do it in an another PR)

src/components/NavBar/store.js
amazingandyyy: replied above

src/components/NavBar/store.js
amazingandyyy: I just try to narrow the scope of the refactor PR.

src/components/NavBar/store.js
alex-cory: That is a great question. I believe they were just there before. Remove and if there are problems, we can tackle them then

src/components/NavBar/index.js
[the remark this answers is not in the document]
henryzbw: I think deric means add observer to `Wallet`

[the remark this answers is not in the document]
amazingandyyy: yea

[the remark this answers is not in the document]
amazingandyyy: I added to Wallet and MyActivities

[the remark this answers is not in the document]
amazingandyyy: thanks

henryzbw:
wait for deric to response his comments

src/components/NavBar/store.js
amazingandyyy: I think moving into GlobalHub is a good idea, updated

src/components/NavBar/index.js
```diff
@@ -37,12 +36,11 @@ const messages = defineMessages({
 @withStyles(sportStyles)
 @withStyles(styles, { withTheme: true })
 @injectIntl
-@connect((state) => ({
-  ...state.App.toJS(),
-  actionableItemCount: state.Graphql.get('actionableItemCount'),
-}))
 @inject('store')
 export default class NavBar extends Component {
+  componentWillMount() {
```
henryzbw: `componentWillMount` will deprecate, use `componentDidMount` instead

amazingandyyy: updated

src/components/NavBar/store.js
dwalintukan: @amazingandyyy It would be a DB problem if there are duplicates, which we handle. But, we should assume that by the time it hits the UI that there wont be duplicates and thus no need to filter out with `uniqBy`. We basically assert that there won't be duplicates when we pull the data from the BE.

It's like saying, why don't we do undefined checks for every value coming from the BE even though we know they will be defined.

src/components/NavBar/store.js
dwalintukan: Let's remove all the uniqBy and map to Oracles since they are not used.

src/redux/Graphql/saga.js
```diff
@@ -23,14 +23,14 @@ import {
...
-const { Pending } = TransactionStatus;
+const { PENDING } = TransactionStatus;
 const { RESULT_SETTING, FINALIZING } = Phases;
 
 const messages = defineMessages({
   placeBet: { id: 'bottomButtonText.placeBet', defaultMessage: 'Place Bet' },
   setResult: { id: 'str.setResult', defaultMessage: 'Set Result' },
   arbitrate: { id: 'bottomButtonText.arbitrate', defaultMessage: 'Arbitrate' },
-  pending: { id: 'str.pending', defaultMessage: 'Pending' },
+  pending: { id: 'str.pending', defaultMessage: 'PENDING' },
```
dwalintukan: assuming this got changed accidentally while `Replacing`

PR #783, opened 2018-07-18
657 createEvent-collabAlex

ippudo:
Um I'm not sure that it seems like I didn't get a cleanup pull


ippudo:
Something happen with oracle branch, which seems to make this branch weird. Will check back later.

PR #786, opened 2018-07-18
657 createEvent-633base

ippudo:
Alternative version based on 663


henryzbw:
why have so many conflicts and commits not yours? Did you branch off from 663-mobx-oracle?

PR #777, opened 2018-07-13
MobX for wallet withdraw

amazingandyyy:

PR #792, opened 2018-07-20
Revert "MobX for wallet withdraw"

amazingandyyy:
Reverts bodhiproject/bodhi-ui#777


amazingandyyy:
The withdraw in topic is broken, revert the change in order to release without new bugs

PR #793, opened 2018-07-20
663 mobx oracle

alex-cory:
- the wider width for the options was approved by derric


src/languageProvider/locales/ko-KR.json
```diff
@@ -42,7 +44,7 @@
   "create.cancel": "취소",
   "create.datePast": "지난 날짜는 선택할 수 없습니다.",
   "create.dateTime": "날짜 및 시간 선택하기",
-  "create.dialogTitle": "이벤트 만들기",
+  "create.dialogTitle": "Create an event",
```
amazingandyyy: we need Korean, need to run [scripts](https://github.com/bodhiproject/bodhi-ui#localization)

src/languageProvider/locales/zh-Hans.json
```diff
@@ -42,7 +44,7 @@
   "create.cancel": "取消",
   "create.datePast": "时间不能为过去",
   "create.dateTime": "选择日期和时间",
-  "create.dialogTitle": "创建事件",
+  "create.dialogTitle": "Create an event",
```
amazingandyyy: as above

src/scenes/App/router.js
```diff
@@ -32,7 +32,7 @@ const AppRouter = ({ url }) => {
       <Route
         exact
         path={`${url}/oracle/:topicAddress/:address/:txid`}
-        component={asyncComponent(() => import('../Event/scenes/oracle'))}
+        component={asyncComponent(() => import('../Event/scenes/Oracle/index'))}
```
amazingandyyy: come on, remove the index

src/scenes/Event/components/EventTxHistory/TxRow.js
```diff
@@ -26,31 +24,17 @@ export default class TxRow extends Component {
   }
 
   toggle = () => {
-    // probably need some logic here for not expanding when clicking on
+    // TODO: probably need some logic here for not expanding when clicking on
```
amazingandyyy: create new issue, I think it deserves a issue

src/scenes/Event/scenes/Oracle/FinalizingOracle.js
```diff
@@ -0,0 +1,49 @@
+/* eslint-disable */
```
amazingandyyy: ughrrrr

src/scenes/Event/scenes/Oracle/Sidebar.js
```diff
@@ -0,0 +1,66 @@
+/* eslint-disable */
+import React from 'react';
```
amazingandyyy: old style:
screen shot 2018-07-20 at 17 19 44

the one in PR:
screen shot 2018-07-20 at 17 20 10


when moving to styled-components, don't sacrafice the styles, it's hard to notice and come back to fix >.<

src/scenes/Event/scenes/Oracle/VotingOracle.js
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable */
```
amazingandyyy: arrrrhh

src/components/StepperVertRight/index.js
```diff
@@ -18,44 +14,17 @@ const POS_BETTING = 1;
 const POS_ORACLE_RESULT_SETTING = 2;
 const POS_OPEN_RESULT_SETTING = 3;
 
-const messages = defineMessages({
-  block: {
```
henryzbw: It's better to keep this. otherwise the script cannot catch the localization string.

src/components/TxSentDialog/index.js
```diff
@@ -0,0 +1,31 @@
...
+
+import styles from './styles';
+
+
+const TxSentDialog = ({ txid, open, onClose, intl, classes }) => (
+  <Dialog open={open} onClose={onClose}>
+    <DialogTitle>{intl.formatMessage({ id: 'transactionSentDialog.successMsg' })}</DialogTitle>
+    <DialogContent>
+      <Typography variant="body1" className={classes.bodyPrimary}>{intl.formatMessage({ id: 'transactionSentDialog.waitingMsg' })}</Typography>
+      <Typography variant="body1">{`${intl.formatMessage({ id: 'str.transactionId' })}: ${txid}`}</Typography>
```
henryzbw: If you do this without defineMessages, I think it may lead to miss some strings in the future. Especially when we want to add more languages.

src/scenes/Event/components/EventTxHistory/TxRow.js
alex-cory: I think it already has an issue. I think Harrison is actually working on this right now

src/components/TxSentDialog/index.js
alex-cory: Are you sure? I just read that all it does is get the language object. [see here](https://github.com/yahoo/react-intl/wiki/API#definemessages)

src/components/StepperVertRight/index.js
alex-cory: Are you sure? I just read that all it does is get the language object. [see here](https://github.com/yahoo/react-intl/wiki/API#definemessages)

henryzbw: I mean 'npm run build:langs', this cannot extract the strings to language files.

alex-cory: why don't we come back to this later so we can keep moving

henryzbw: created an issue. You can keep moving. You can fix it after merge.

PR #798, opened 2018-07-24
Mobx create event

alex-cory:
BOOM 🔥
things that need clarification:

- [x] validation for title, do we need that web3utils stuff?
- [x] in `prepareToCreateEvent` in `CreateEvent/store.js`, how do you want to handle the `if (wallet.needsToBeUnlocked)` stuff? (solution: have andy add it when he finishes wallet)
- [x] I think we should do optimistic updates when submitting. See comment in the `submit` method
- [ ] I think the sortby is backwards for qtumPrediction


henryzbw:
image
style error.
and when scroll like this, the remove button cannot be clicked.

henryzbw:
image
When add more outcomes, sometime will throw error

henryzbw:
Prediction Start Time, Prediction End Time, Result Setting Start Time, Result Setting End Time
need to be in increasing time, now, it can be anytime event if end before start.
image

src/scenes/CreateEvent/store.js
```diff
@@ -0,0 +1,397 @@
...
+      && this.resultSetter
+      // no errors
+      && !title
+      && !creator
+      && !prediction.startTime
+      && !prediction.endTime
+      && !resultSetting.startTime
+      && !resultSetting.endTime
+      && outcomes.every(outcome => !outcome)
+      && !resultSetter
```
dwalintukan: arent these checking for undefined instead of empty?

src/scenes/CreateEvent/store.js
```diff
@@ -0,0 +1,397 @@
...
+      // console.log('CREATE EVENT TX: ', data.createTopic);
+      runInAction(() => {
+        // optimistically add the oracle to the qtum prediction page list
+        // this.app.qtumPrediction.list.push(new Oracle(data.createTopic));
+        this.txConfirmDialogOpen = false;
+        this.txid = data.createTopic.txid;
+        this.txSentDialogOpen = true;
+      });
+    } catch (error) {
+      console.error('ERROR: ', { // eslint-disable-line
```
dwalintukan: add TODO to send this to ErrorDialog

src/scenes/CreateEvent/store.js
```diff
@@ -0,0 +1,397 @@
...
+        this.outcomes,
+        this.resultSetter,
+        moment(this.prediction.startTime).utc().unix().toString(),
+        moment(this.prediction.endTime).utc().unix().toString(),
+        moment(this.resultSetting.startTime).utc().unix().toString(),
+        moment(this.resultSetting.endTime).utc().unix().toString(),
+        decimalToSatoshi(this.escrowAmount),
+        this.creator, // address
+      );
+      // console.log('CREATE EVENT TX: ', data.createTopic);
```
dwalintukan: remove

src/scenes/CreateEvent/store.js
```diff
@@ -0,0 +1,397 @@
...
+      runInAction(() => {
+        // optimistically add the oracle to the qtum prediction page list
+        // this.app.qtumPrediction.list.push(new Oracle(data.createTopic));
```
dwalintukan: does this mean the updating does not work currently?

src/scenes/Event/scenes/Oracle/components/Option/index.js~HEAD
```diff
@@ -0,0 +1,156 @@
+import React, { Component, Fragment } from 'react';
```
dwalintukan: looks like a rebase error. assuming this class got deleted before.

src/scenes/Event/scenes/Oracle/components/Option/styles.js~HEAD
```diff
@@ -0,0 +1,79 @@
+const styles = (theme) => ({
```
dwalintukan: same. needs deleting?

src/scenes/CreateEvent/store.js
alex-cory:
```js
!'' === true
!undefined === true
!null === true
```
since all these are strings, we're just checking to see if they are empty, if they are, we're good

src/scenes/CreateEvent/store.js
alex-cory: yes

alex-cory: but now it's fixed

PR #817, opened 2018-08-14
357 datepicker version2

ippudo:
Mixed components from **Material ui pickers** and legacy **Material ui**'s pickers


ippudo:
Now CreateEvent Store use unix stamp to store time, instead of the "YYYY/MM/DDTHH:mm" formatted string.

PR #816, opened 2018-08-13
Update Prediction for a demo of make store, topActions locally. + update Card design

amazingandyyy:
This PR change codebase related to QtumPrediction
1. move store/QtumPrediciton to scene/QtumPrediciton/store.js, prepare for testing
2. update Sorting Method and prepare for more sorting options
3. change the info in EventCard and also move EventCard to local

what this PR is lack of is 
1. not enough sorting options

I am still updating it.

PR #845, opened 2018-08-14
Revert "Update Prediction for a demo of make store, topActions locally. + update Card design"

amazingandyyy:
Reverts bodhiproject/bodhi-ui#816


henryzbw:
why revert?

amazingandyyy:
cause andy is dumb

PR #846, opened 2018-08-14
804 localize invalid option

BernardHan:


src/scenes/Activities/ActivityHistory/EventRows/index.js
```diff
@@ -63,7 +63,7 @@ class EventRow extends Component {
             <TableCell className={classes.summaryRowCell}>{getShortLocalDateTimeString(createdTime)}</TableCell>
             <TableCell>{getTxTypeString(type, locale, localeMessages)}</TableCell>
             <NameLinkCell className={classes.viewEventLink} clickable={topic && topic.address} onClick={this.handleClick(txid, topic && topic.address)}>
-              {name || (topic && topic.name)}
+              {name === 'Invalid' && localizedInvalid !== undefined ? localizedInvalid[intl.locale.slice(0, 2)] : name || (topic && topic.name)}
```
dwalintukan: move this logic into model (Oracle) and pass in the locale.

good practice is to keep the UI as dumb as possible. get data and show. in this case, let the model do the heavy lifting.

src/stores/models/Oracle.js
```diff
@@ -123,5 +125,11 @@ export default class Oracle {
     this.endTime = this.phase === RESULT_SETTING ? oracle.resultSetEndTime : oracle.endTime;
 
     this.options = oracle.options.map((option, i) => new Option(option, i, this, app));
+
+    this.localizedInvalid = {
```
dwalintukan: lets make a convenience getter method for getting the array of options. something like:
```
getOptions = (locale) => {
  const localizedInvalid = {
    en: 'Invalid',
    zh: '无效',
    ko: '무효의',
  }
  return _.map(this.options, (option) => {
    if (option === 'Invalid') {
      return localizedInvalid[locale.slice(0, 2)];
    }
    return option;
  });
}
```

haven't tested the above code, but i think its correct.

then just use this method anytime you get the options array.

src/scenes/Activities/ActivityHistory/EventRows/index.js
BernardHan: I actually can use the original code, since you issue this as a bug in sprint board, "activity history name should be event name, not option name"

dwalintukan: Ah yes you are right. This is showing the wrong thing here. Transaction History in the event detail page is where you should show the localized Invalid (if necessary)

src/stores/models/Oracle.js
BernardHan: I made a parse function inside localizedInvalid instead of this mapping since what calls it is already inside a mapping.

dwalintukan:
Please rebase your branch. It is showing commit changes from Henry's testing branch.

PR #848, opened 2018-08-15
830 new oracle crash

dwalintukan:
- Fixed couple bugs (see referenced issues below)
- Cleaned up `EventPage` store
- Rearranged some network files


src/network/graphql/mutations.js
```diff
@@ -1,8 +1,8 @@
 import gql from 'graphql-tag';
 import _ from 'lodash';
 
-import client from './graphClient';
-import { TYPE, getMutation, isValidEnum } from './graphSchema';
+import client from '.';
```
amazingandyyy: `import client from './index';` was surprised to see that lol

dwalintukan: oh it used to be called `graphClient`

Issue #835, opened 2018-08-14
Result History showing current round


dwalintukan:
This event is in Arbitration Round 1. It should not be showing the current round in the `Result History`. Believe this is a bug.

![image](https://user-images.githubusercontent.com/4350404/44092609-1d4e07c6-9ffb-11e8-9d76-1921e2d8a8e3.png)

ippudo:
Is this really consider to a bug? I think the whole logic at result history is weird and need to be discussed

PR #853, opened 2018-08-15
835-resultHistoryVote

ippudo:
#835


ippudo:
TBH the whole part code here looks weird

src/scenes/Event/components/ResultHistory/index.js
```diff
@@ -76,6 +76,7 @@ export default class EventResultHistory extends Component {
             <TableBody>
               {_.map(sortedOracles, (oracle, index) => {
                 const invalidOption = localizeInvalidOption(oracle.options[oracle.resultIdx], intl);
+                if (currentEvent.phase === Phases.VOTING && index === sortedOracles.length - 1) return;
```
dwalintukan: Lets remove the Oracle object from `sortedOracles` for Voting phase instead of not returning the the `TableRow`

ippudo: Tried simply filter Voting phase at sortedOracles but failed. It will affect other oracle pages because of the magic codes here.

PR #867, opened 2018-08-17
update codes with Harrison for the new block transactions bugs

### LINKED ISSUE #866: Appolo error at console while block number updated
appolo

This bugs doesn't related to removing redux.

amazingandyyy:
fix #866

### DIFF
--- a/src/scenes/Event/store.js
+++ b/src/scenes/Event/store.js
@@ -29,6 +29,7 @@ const INIT = {
   warningType: '',
   eventWarningMessageId: '',
   escrowClaim: 0,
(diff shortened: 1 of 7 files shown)


src/scenes/Event/store.js
```diff
@@ -131,6 +131,7 @@ export default class EventStore {
   @action
   initTopic = async () => {
     // GraphQL calls
+    this.app.ui.location = Routes.TOPIC;
```
ippudo: Editing this will leads backbutton boom. Need other solution.

PR #874, opened 2018-08-19
821 fix todos

dwalintukan:


BernardHan:
When staying in unconfirmed event page, when the event gets confirmed, the page does not get refreshed. Not sure if this is intentional or not.

src/scenes/Event/components/ResultHistory/index.js
```diff
@@ -74,33 +81,38 @@ export default class EventResultHistory extends Component {
...
-                    <TableCell padding="dense">{`${_.sum(oracle.amounts)} ${oracle.token}`}</TableCell>
-                  </TableRow>
-                );
-              })}
+              <ResultRows sortedOracles={sortedOracles} getTypeText={this.getTypeText} {...this.props} />
             </TableBody>
           </Table>
-        ) : (
-          <Typography variant="body1">
-            <FormattedMessage id="str.emptyTxHistory" defaultMessage="You do not have any transactions right now." />
```
henryzbw: Do we need to have this no transactions part ?

src/scenes/Event/components/TransactionHistory/TxRow.js
```diff
@@ -72,23 +69,28 @@ export default class TxRow extends Component {
 
     return (
       <Fragment>
-        <TableRow key={`tx-${txid}`} onClick={this.toggle}>
+        <TableRow key={`tx-${txid}`}>
```
henryzbw: don't we support click the row to expand?

src/scenes/Event/store.js
```diff
@@ -459,30 +469,11 @@ export default class EventStore {
       return;
     }
 
-    // Trying to vote over the consensus threshold
-    const optionAmount = this.selectedOption.amount;
```
henryzbw: Now we support vote over threshold?

src/scenes/Wallet/Balances/index.js
```diff
@@ -384,21 +383,17 @@ export default class MyBalances extends Component {
   };
 
   onWithdrawClicked(event) {
-    const { wallet, walletUnlockDialog } = this.props.store;
```
henryzbw: we don't need to show walletUnlock?

src/scenes/Event/components/ResultHistory/index.js
dwalintukan: Decided to just remove it entirely.
1) it was the wrong text.
2) thought it wasn't relevant to show this message for the user. thought it would just confuse them more.
3. It is a bug if they see this message since we intentionally start showing the result history from the first arbitration round and on.

src/scenes/Event/components/TransactionHistory/TxRow.js
dwalintukan: I moved the click to the arrow icon only because it felt like bad UI when I wanted to copy and address or some other text, it would keep toggling the expansion.

src/scenes/Event/store.js
dwalintukan: This code is unreachable now. Since we auto adjust the `amount` field to the maximum of the threshold, it will never hit this block.

src/scenes/Wallet/Balances/index.js
dwalintukan: Showing the wallet unlock is just a fallback now. Decided to not spend the extra time to implement this properly since:
1. We are going light client and using qrypto
2. even for the future Electron version, we unlock at the very beginning before they even load the UI

PR #889, opened 2018-08-21
878 insight path

dwalintukan:
- Fix insight api path
- Use `NODE_ENV` to determine testnet/mainnet env


dwalintukan:
@ippudo So far I have been running the dev version only remotely. Let me think about the switching for now. But we use the `NODE_ENV` var to switch environments.

PR #890, opened 2018-08-21
876-tooltips

ippudo:


src/scenes/Event/WithdrawingTopic/Reward.js
```diff
@@ -16,52 +17,56 @@ const messages = defineMessages({
   },
 });
 
-// TODO: fix the tooltip next to QtumReturn and BotUsed
-const Reward = ({ eventPage, topic }) => {
```
dwalintukan: FYI you can inject styles inline with stateless components.

Issue #789, opened 2018-07-18
Checking QTUM amount of wrong total in [CreateEvent]


amazingandyyy:
Checking QTUM amount of wrong total in [CreateEvent]
And also need to check BOT amount

amazingandyyy:
related to #656

ippudo:
BOT amount is validated by another standalone function.
But I'm feeling the total amount and display are wrong.
Need further check.

ippudo:
Part done, other go #894

Issue #788, opened 2018-07-18
Checking QTUM amount of wrong total in [BET]


amazingandyyy:
In bet, qtum amount check on option select is based on the total of all address, but it should be based on the address you select

ippudo:
And also for withdrawal dialog I believe I saw such similar codes. Lots work to do lol

PR #893, opened 2018-08-22
788 wrong qtum amount validation for BET and WITHDRAWDIALOG

### LINKED ISSUE #790: Should remove the `token amount not enough` warning when changing wallet [Oracle]
In oracle detail page, once detect token amount not enough, and then change wallet address, need to enter a number to remove the `token amount not enough` warning

ippudo:
NOTE: **NOT FULLY TESTED**. Need more people to try.

#788 , both BET in the topic and WITHDRAWDIALOG in the followup.
Part of #789 since CREATEEVENT also use wrong amount of QTUM, but I'm feeling problems at 789 is more than this.

Possibly fixed #790 but not able to test

### DIFF
--- a/src/scenes/Event/store.js
+++ b/src/scenes/Event/store.js
@@ -190,7 +190,7 @@ export default class EventStore {
 
     // Toggle CTA on new block, transaction change, amount input change, option selected
     reaction(
-      () => this.app.global.syncBlockTime + this.transactions + this.amount + this.selectedOptionIdx,
+      () => this.app.global.syncBlockTime + this.transactions + this.amount + this.selectedOptionIdx + this.app.wallet.lastUsedAddress,
       () => {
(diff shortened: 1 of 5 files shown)


src/scenes/Event/store.js
```diff
@@ -383,8 +383,8 @@ export default class EventStore {
     const { phase, resultSetterQAddress, resultSetStartTime, isOpenResultSetting, consensusThreshold } = this.oracle;
     const { global: { syncBlockTime }, wallet } = this.app;
     const currBlockTime = moment.unix(syncBlockTime);
-    const totalQtum = _.sumBy(wallet.addresses, ({ qtum }) => qtum);
-    const notEnoughQtum = totalQtum < maxTransactionFee;
+    const currentWalletQtum = wallet.lastUsedWallet.qtum;
```
amazingandyyy: `wallet.lastUsedWallet` looks cool

Issue #894, opened 2018-08-22
Wrong BOT amount shown at Create Event confirm dialog


ippudo:
![create1](https://user-images.githubusercontent.com/9134404/44437990-f1b72800-a571-11e8-8b37-80669b6fe9c9.png)

The data from the post API returns array of 2 and each has a BOT amount.

ippudo:
By the way if you has already got an successful "Approve" transaction but "create event" transaction is still pending, then you will get a different txfees data from the API.
![create2](https://user-images.githubusercontent.com/9134404/44438049-393db400-a572-11e8-9368-a32535985ce6.png)
Though it has no meaning since you can't create any new event at that period, and will get this:
![create3](https://user-images.githubusercontent.com/9134404/44438115-8c176b80-a572-11e8-96dd-98cdde377aef.png)

dwalintukan:
Ya that logic with the `Approve` is expected. We should better inform the user that they cant create another Event until the other one is finished.

1. Disable Create button
2. Add an EventWarning that there is a pending Create Event tx

Issue #892, opened 2018-08-21
Outdated maxTransactionFee


ippudo:
We're now setting maxTransactionFee to 0.1 for validation use, but the transaction fee can easily beyond this amount ( around 0.27 in my Computer ).

![create4](https://user-images.githubusercontent.com/9134404/44437963-c6343d80-a571-11e8-9b81-56eac814edd8.png)
Then we'll see a lot of this by doing BET/CREATE/TRANSFER when we're low in qtum

dwalintukan:
@ippudo Lets remove this static `0.1 maxTransactionFee`. We can replace it with the actual max fee with the `/transaction-cost` api.

Issue #847, opened 2018-08-15
Move EventWarning above the CTA button and grey CTA button if disabled in Event detail


dwalintukan:
- [ ] Move the event warning (light blue box with text) right above the CTA button
- [x] Grey out the CTA button if it is disabled

![image](https://user-images.githubusercontent.com/4350404/44139580-a7de5ef8-a0a1-11e8-84f1-44baa5edfae5.png)

dwalintukan:
Ok, we can discuss this at the next meeting.

ippudo:
Move the event warning from the light blue box and fill them into error for each fields.

dwalintukan:
@ippudo Actually, its:
1. Keep any of the `EventWarningType.INFO or HIGHLIGHT` in the same spot
2. Add warning text under the `Amount` and `Address` fields. Don't use the same EventWarning component. Just some small text will do, ask Lili. Pretty sure she will follow the same design as Qrypto.

PR #906, opened 2018-08-23
894 create event

ippudo:
using API to detect the max transaction fee for Create Event ( #892 )
remove bot shown on "approve" at txconfirmdialog  ( #894 )
correct css style for button on oracle pages ( #847 )

### DIFF
--- a/src/scenes/CreateEvent/store.js
+++ b/src/scenes/CreateEvent/store.js
@@ -172,32 +172,6 @@ export default class CreateEventStore {
     );
   }
 
-  @action
-  onEnterDialog = async () => {
-    try {
-      const { data: { result } } = await axios.post(
-        Routes.api.transactionCost,
-        {
-          type: TransactionType.APPROVE_CREATE_EVENT,
-          token: Token.BOT,
-          amount: decimalToSatoshi(this.escrowAmount),
-          optionIdx: undefined,
-          topicAddress: undefined,
-          oracleAddress: undefined,
-          senderAddress: this.app.wallet.lastUsedAddress,
-        }
-      );
-      const txFees = _.map(result, (item) => new TransactionCost(item));
-      runInAction(() => {
-        this.txFees = txFees;
-      });
-    } catch (error) {
-      runInAction(() => {
-        this.app.ui.setError(error.message, Routes.api.transactionCost);
-      });
-    }
-  }
-
   constructor(app) {
     this.app = app;
     reaction( // when we add the creator, update our last used address
@@ -283,7 +257,7 @@ export default class CreateEventStore {
       this.averageBlockTime = defaults.averageBlockTime;
     }
 
-    runInAction(() => {
+    runInAction(async () => {
       this.prediction.startTime = nowPlus(TIME_DELAY_FROM_NOW_SEC);
       this.prediction.endTime = nowPlus(TIME_DELAY_FROM_NOW_SEC + TIME_GAP_MIN_SEC);
       this.resultSetting.startTime = nowPlus(TIME_DELAY_FROM_NOW_SEC + TIME_GAP_MIN_SEC);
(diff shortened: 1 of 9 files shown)


src/scenes/CreateEvent/store.js
```diff
@@ -171,6 +177,32 @@ export default class CreateEventStore {
...
+    try {
+      const { data: { result } } = await axios.post(
+        Routes.api.transactionCost,
+        {
+          type: TransactionType.APPROVE_CREATE_EVENT,
+          token: Token.BOT,
+          amount: decimalToSatoshi(this.escrowAmount),
+          optionIdx: undefined,
+          topicAddress: undefined,
+          oracleAddress: undefined,
```
dwalintukan: Don't need to add any `undefined` keys, just leave them out. By default, every var not explicitly defined is `undefined`.

src/scenes/Event/store.js
```diff
@@ -118,6 +118,30 @@ export default class EventStore {
     this.setReactions();
   }
 
+  @action
+  async initTxfees() {
```
dwalintukan: Why do you need this? Tx cost should be calculated when the user does the action. And this one is assuming we are doing a `bet`, but the EventStore is used for all types of txs. 

It is also unused, remove it.

src/scenes/CreateEvent/store.js
```diff
@@ -171,6 +177,32 @@ export default class CreateEventStore {
...
+          senderAddress: this.app.wallet.lastUsedAddress,
+        }
+      );
+      const txFees = _.map(result, (item) => new TransactionCost(item));
+      runInAction(() => {
```
dwalintukan: Don't believe you need to wrap this in an action. It's already in an action.

src/scenes/CreateEvent/store.js
```diff
@@ -171,6 +177,32 @@ export default class CreateEventStore {
...
+        this.txFees = txFees;
+      });
+    } catch (error) {
+      runInAction(() => {
+        this.app.ui.setError(error.message, Routes.api.transactionCost);
```
dwalintukan: same, prob dont need to wrap in `runInAction`

src/scenes/CreateEvent/store.js
```diff
@@ -171,6 +177,32 @@ export default class CreateEventStore {
     );
   }
 
+  @action
+  onEnterDialog = async () => {
+    try {
+      const { data: { result } } = await axios.post(
```
dwalintukan: Why is this logic duplicated in two places? it is the same logic as `prepareToCreateEvent()`. We don't need to call this logic twice.

dwalintukan:
@ippudo Also, for this #847 you forgot to handle number 1.

src/scenes/Event/components/Button.js
```diff
@@ -6,8 +6,10 @@ export const Button = withStyles(styles, { withTheme: true })(({ classes, ...pro
   <_Button
     fullWidth
     size='large'
+    color='default'
```
dwalintukan: why default? we normally use `primary` to use the defined primary color from the theme.

src/scenes/Event/components/Button.js
```diff
@@ -6,8 +6,10 @@ export const Button = withStyles(styles, { withTheme: true })(({ classes, ...pro
...
+    variant='outlined'
```
dwalintukan: whats the difference of `outlined` and `raised`?

src/scenes/Event/components/styles.js
```diff
@@ -1,11 +1,8 @@
 const styles = (theme) => ({
   oracleButton: {
+    marginTop: theme.padding.md.px,
+    background: theme.palette.primary.main,
     color: theme.palette.primary.contrastText,
```
dwalintukan: if you set `color='primary'` in the button itself, you shouldn't have to define the colors. look at how we use all the other buttons, eg. `Wallet/index.js L302`. when it comes to UI elements, try to be consistent with the same implementation as other places.

dwalintukan:
@ippudo Don't forget to handle number 1 in #847.

ippudo:
Some UI changes that not mentioned at the issues are expected with this pr.

PR #927, opened 2018-09-05
test cases complete

BernardHan:
#914

### DIFF
--- a/src/network/graphql/__mocks__/queries.js
+++ b/src/network/graphql/__mocks__/queries.js
@@ -1,0 +1,36 @@
+/** This is a mock module of ../queries.js,
+ * so the test files link to this instead of connecting to backend
+ * This file simply skip the backend retrival and fake query results and return
+ * Current logic is straightforward but good enough to make tests
+ * */
+export function queryAllTopics(filters, orderBy, limit, skip) {
+  const topics = [];
+  const max = 23;
+  let i;
+  for (i = 0; i < max; i++) {
+    topics.push({ txid: i });
+  }
+  const end = skip + limit <= max ? skip + limit : max;
+  return topics.slice(skip, end);
+}
+
+export function queryAllOracles(filters, orderBy, limit, skip) {
+  const oracles = [];
+  const max = 23;
+  let i;
+  for (i = 0; i < max; i++) {
+    const oracle = {
+      txid: i,
+      amounts: [],
+      consensusThreshold: 100,
+      address: '02e91962156da21fae38e65038279c020347e4ff',
+      topicAddress: '4044f951857f2885d66d29a475235dacdaddea84',
+      resultSetEndTime: 10,
+      endTime: 20,
+      options: [],
+    };
+    oracles.push(oracle);
+  }
+  const end = skip + limit <= max ? skip + limit : max;
+  return oracles.slice(skip, end);
+}
--- a/src/stores/AllEventsStore.test.js
+++ b/src/stores/AllEventsStore.test.js
@@ -14,9 +14,14 @@ global.localStorage = localStorageMock;
 
 describe('AllEventsStore', () => {
   let store;
+  const addr = {
+    address: 'qSu4uU8MGp2Ya6j9kQZAtizUfC82aCvGT1',
+    qtum: 2000,
+    bot: 100,
+  };
   const app = {
     sortBy: 'ASC',
-    wallet: { addresses: 0 },
+    wallet: { addresses: [addr] },
     global: { syncBlockNum: 0 },
     refreshing: false,
     ui: { location: 0 },
@@ -64,7 +69,12 @@ describe('AllEventsStore', () => {
   it('Reaction Wallet Addr', () => {
     expect(store.list.length).toBe(0);
     app.ui.location = Routes.ALL_EVENTS;
-    app.wallet.addresses = 1;
+    const newAddr = {
+      address: 'qSu4uU8MGp2Ya6j9kQZAtizUfC82aCvGT2',
+      qtum: 200,
+      bot: 10,
+    };
+    app.wallet.addresses.push(newAddr);
     setTimeout(() => {
       expect.assertions(1);
       expect(store.list.length).toBe(store.limit);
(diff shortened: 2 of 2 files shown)


src/network/graphql/__mocks__/queries.js
```diff
@@ -0,0 +1,36 @@
+/** This is a mock module of ../queries.js,
```
dwalintukan: can we rename the folder for `__mocks__` to just `mocks`? or is it needed to have the underscores?

src/stores/AllEventsStore.test.js
```diff
@@ -0,0 +1,91 @@
+import { Routes } from 'constants';
+jest.mock('../network/graphql/queries'); // block and manually mock our backend
```
dwalintukan: move under imports?

src/stores/AllEventsStore.test.js
```diff
@@ -0,0 +1,91 @@
...
+const localStorageMock = {
+  getItem: jest.fn(),
+  setItem: jest.fn(),
+  clear: jest.fn(),
+};
+const navigatorMock = {};
+global.navigator = navigatorMock;
+global.localStorage = localStorageMock;
+
+/** test */
```
dwalintukan: remove unnecessary comments

src/stores/AllEventsStore.test.js
```diff
@@ -0,0 +1,91 @@
...
+  }; // mock the appstore
+  beforeEach(() => {
+    store = new AllEventsStore(app); // create a new instance before each test case
+  });
+
+  /** all following test cases target specifically to the mock backend, eg. network/graphql/__mocks__/queries.js */
+  it('Init', async () => {
+    const curListNum = store.list.length;
+    expect(curListNum).toBe(0);
+    expect.assertions(4);
```
dwalintukan: curious, what does this do?

src/stores/AllEventsStore.test.js
```diff
@@ -0,0 +1,91 @@
...
+    await store.init();
+    expect(app.ui.location).toBe(Routes.ALL_EVENTS); // test webpage location
+    expect(store.list.length).toBe(curListNum + store.limit); // test if list got new cards
+    expect(store.loading).toBe(false); // test runInAction changes
```
dwalintukan: can remove unnecessary comments. these are self-explanatory by the `expect.toBe` as they read like sentences

src/stores/AllEventsStore.test.js
```diff
@@ -0,0 +1,91 @@
...
+    expect(store.list.length).toBe(store.limit);
+    expect(store.loadingMore).toBe(false);
+    await store.loadMoreEvents();
+    expect(store.list.length).toBe((curListNum + store.limit) - 2); // mock backend only has 46 events in total
+    expect(store.loadingMore).toBe(false);
+    expect(store.hasMore).toBe(false); // no events left in backend
+    expect(store.skip).toBe(curSkip + store.limit);
+  });
+
+  /** following test cases test the reaction function */
```
dwalintukan: again, unnecessary since we can tell by the test name

src/stores/AllEventsStore.test.js
```diff
@@ -0,0 +1,91 @@
...
+  it('Reaction SortBy', () => {
+    app.ui.location = Routes.ALL_EVENTS;
```
dwalintukan: would be good to check the starting state of the list length before the reaction

src/stores/AllEventsStore.test.js
```diff
@@ -0,0 +1,91 @@
...
+    setTimeout(() => {
+      expect.assertions(2);
+      expect(store.app.sortBy).toBe('Yo');
+      expect(store.list.length).toBe(store.limit);
+    }, 500);
+  });
+
+  it('Reaction Wallet Addr', () => {
+    app.ui.location = Routes.ALL_EVENTS;
+    app.wallet.addresses = 1;
```
dwalintukan: when testing the reaction, we should test with how a real change should be. this should be an array of objects.

src/stores/AllEventsStore.test.js
BernardHan: Initially I thought this has to be before the import module that uses backend, but now it seems unnecessary. Moved.

src/network/graphql/__mocks__/queries.js
BernardHan: It seems to be a must. Google & Jest official tells me nothing about changing the folder name.

src/stores/AllEventsStore.test.js
BernardHan: Line 31 & 32 is trying to make sure the list is empty before init.
Line 33 says this test case should pass 4 assertions. This line is unnecessary for this test case but good to have. But in general, it's useful in asynchronous tests.
For example,
`it('calls the API and throws an error', async () => {
  expect.assertions(2);
  try {
    await login('email', 'password');
  } catch (error) {
    expect(error.name).toEqual('Unauthorized');
    expect(error.status).toEqual(401);
  }
});`
Without it, this test case will pass even when login successful.

Or 
`test('doAsync calls both callbacks', () => {
  expect.assertions(2);
  function callback1(data) {
    expect(data).toBeTruthy();
  }
  function callback2(data) {
    expect(data).toBeTruthy();
  }

  doAsync(callback1, callback2);
});`
Here it ensures that both callbacks are actually called.

src/stores/AllEventsStore.test.js
BernardHan: removed

src/stores/AllEventsStore.test.js
BernardHan: Are you saying I need to check the list retrieved from the mocked backend should be the same as what I stuff in the mocked backend?

src/network/graphql/__mocks__/queries.js
dwalintukan: does the tests work if you do rename it?

src/stores/AllEventsStore.test.js
dwalintukan: gotcha, its just saying how many asserts are required.

src/network/graphql/__mocks__/queries.js
BernardHan: no it does not

Issue #873, opened 2018-08-19
Implement fetching Oracle by `hashId` when implemented in backend


dwalintukan:
https://github.com/bodhiproject/bodhi-server/issues/77

1. create Unconfirmed route -> Event detail page
2. handle on new block num
3. route to confirmed Event Detail page if Oracle is confirmed

PR #946, opened 2018-09-11
hashId integrated to let unconfirmed to confirmed automatically

henryzbw:
#873

### DIFF
--- a/src/scenes/Event/store.js
+++ b/src/scenes/Event/store.js
@@ -206,9 +206,8 @@ export default class EventStore {
   @action
   verifyConfirmedOracle = async () => {
     const res = await queryAllOracles([{ hashId: this.hashId }]);
-    const confirmedOracle = _.filter(res, 'topicAddress');
-    if (!_.isEmpty(confirmedOracle)) {
-      const { topicAddress, address, txid } = confirmedOracle[0];
+    if (!_.isNull(res[0].topicAddress)) {
+      const { topicAddress, address, txid } = res[0];
       this.app.router.push(`/oracle/${topicAddress}/${address}/${txid}`);
     }
   }
--- a/src/scenes/Event/store.js
+++ b/src/scenes/Event/store.js
@@ -31,6 +31,7 @@ const INIT = {
   warningType: '',
   eventWarningMessageId: '',
   escrowClaim: 0,
+  hashId: '',
 };
 
 export default class EventStore {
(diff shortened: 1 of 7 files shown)


src/scenes/Event/store.js
```diff
@@ -72,7 +74,7 @@ export default class EventStore {
   }
   // For Oracle only
   @computed get unconfirmed() {
-    return this.topicAddress === 'null' && this.address === 'null';
+    return _.isUndefined(this.topicAddress) && _.isUndefined(this.address);
```
dwalintukan: are you sure this works? I believe the `topicAddress` and `address` are actually a string that contains `null`

src/scenes/Event/store.js
```diff
@@ -200,12 +205,12 @@ export default class EventStore {
...
   @action
   verifyConfirmedOracle = async () => {
-    // TODO: need to implement hashedId on server first so we can fetch the updated Oracle by hashedId as it goes through the approve -> createEvent txs
-    // const res = await queryAllOracles([{ txid: this.txid }]);
-    // const { topicAddress, address, txid } = res[0];
-    // if (address && topicAddress) {
-    //   this.app.router.push(`/oracle/${topicAddress}/${address}/${txid}`);
-    // }
+    const res = await queryAllOracles([{ hashId: this.hashId }]);
+    const confirmedOracle = _.filter(res, 'topicAddress');
```
dwalintukan: why do we need to filter? there should only be 1 returning?

src/scenes/Event/store.js
henryzbw: we used to have  `null` in our path, but now we don't have it, so the destruct of `topicAddress` and `address` cannot get `null`

src/scenes/Event/store.js
henryzbw: this is used to handle the `approve` tx, since this time the oracle doesn't have topicAddress in it, so we should not push the new path.

src/scenes/Event/store.js
dwalintukan: oh right, duhhhh lol

src/scenes/Event/store.js
dwalintukan: hm? I'm not understanding this. The `queryAllOracles` res should have only 1 Oracle obj in the array. So we don't need to `filter` the 1 item array.

PR #948, opened 2018-09-11
Searchbar

ippudo:
With bodhi-server version: https://github.com/bodhiproject/bodhi-server/pull/117
#918


src/network/graphql/queries.js
```diff
@@ -224,3 +238,27 @@ export function querySyncInfo(includeBalance) {
   }
   return request.execute();
 }
+
+/*
+* Search for Oracles that contains phrase either in title or result setter address
+* @param phrase: String
+*/
```
dwalintukan: comments should follow this pattern:
```
// notice the double ** after forward-slash. VSCode should auto complete after /** and pressing enter
/** 
 * The description of the function here.
 * @param {typeOfParam} nameOfParam The description of the param.
 * @return {typeOfReturn} The description of the return.
 */
```

src/scenes/Search/index.js
```diff
@@ -0,0 +1,28 @@
...
+import EventCard from '../../components/EventCard';
+
+@inject('store')
+@observer
+export default class Search extends Component {
+  render() {
+    const { ui } = this.props.store;
+    const { list, loading } = this.props.store.search;
+    const events = (list || []).map((event, i) => (<EventCard onClick={() => ui.disableSearchBarMode()} key={i} index={i} event={event} />));
+    const noResult = (
```
dwalintukan: make this a stateless component

src/components/NavBar/index.js
```diff
@@ -83,42 +103,90 @@ const QAButton = ({ intl, changeDropDownDirection }) => (
...
       <NavBarRightButton>
         <FormattedMessage id="navBar.activities" defaultMessage="My Activities" />
       </NavBarRightButton>
-    </NavLink>
-  </NavBarRightButtonContainer>);
+    </NavBarRightButtonContainer>
+  </NavLink>);
 });
 
+const SearchButton = observer(({ onClick }) => (
```
dwalintukan: don't pass the `onClick` functionality. call the store func directly from the component. you can `inject('store')` to stateless components.

src/components/NavBar/index.js
```diff
@@ -83,42 +103,90 @@ const QAButton = ({ intl, changeDropDownDirection }) => (
...
+  color: rgba(255, 255, 255, 0.65);
+  padding-left: 14px;
+`;
+const DivSearchBarField = styled.div`
+  margin: auto;
+  display: flex;
+  width: 60%;
+`;
+
+const SearchBarField = ({ intl, classes, store: { search }, onClose, onSearchBarKeyDown }) => (
```
dwalintukan: same, dont pass store onClick listeners. just access the store func directly from the component.

dwalintukan:
also needs rebase

ippudo:
There's some new padding values from the new design. Should I add new value to the theme, or modify the design to fit the existing theme padding values?

dwalintukan:
@ippudo Whenever possible, try to adapt the design to fit the existing theme values. eg. 10px padding => 12px theme padding

The theme values shouldnt change anymore unless there is a new value that is commonly used that we dont have in there.

src/components/NavBar/index.js
```diff
@@ -126,7 +127,7 @@ const DivSearchBarField = styled.div`
   width: 60%;
 `;
 
-const SearchBarField = ({ intl, classes, store: { search }, onClose, onSearchBarKeyDown }) => (
+const SearchBarField = ({ intl, classes, store: { search, ui }, onSearchBarKeyDown }) => (
```
dwalintukan: should inject the store like you did with the `SearchButton`. be consistent with your coding practices.

src/network/graphql/queries.js
```diff
@@ -239,10 +239,11 @@ export function querySyncInfo(includeBalance) {
   return request.execute();
 }
 
-/*
-* Search for Oracles that contains phrase either in title or result setter address
-* @param phrase: String
-*/
+/**
+ * Search for Oracles that contains phrase either in title or result setter address
+ * @param {String} Phrase The keyword param for search
```
dwalintukan: the param name should match exactly as the param `phrase` lowercase

src/components/NavBar/index.js
ippudo: Do we also need to do the same way with Withstyles/Intl inject on these components?
And I saw some code at navbar components also using the similar way to access the store, should I also modify them as well?

dwalintukan: @ippudo ya we should be consistent across all stateless components. if we inject store or withStyles in one component, we should do it with the others. don't worry about refactoring all the ones that aren't touched in this PR. can do that later.

Issue #957, opened 2018-09-17
Search: should show loading spinner while the search request is happening


dwalintukan:
Right now the tabs just show empty while a search request is happening. Should show a loading spinner (like the one for loading more) while the request is going.

PR #963, opened 2018-09-17
fix search showing wrong phase

### LINKED ISSUE #956: Search results showing items in wrong phase
This testnet event `Qrypto Dapp Test` is still in the betting phase, but the search results is showing it in the withdraw phase.

image
image

BernardHan:
along with server PR, this fix #956 , also #957

### DIFF
--- a/src/scenes/Search/store.js
+++ b/src/scenes/Search/store.js
@@ -40,6 +40,14 @@ export default class SearchStore {
         }
       },
     );
+    reaction(
+      () => this.app.global.syncBlockNum,
+      () => {
+        if (this.app.ui.searchBarMode && !_.isEmpty(this.phrase)) {
+          this.init();
+        }
+      }
+    );
   }
 
   @action
@@ -86,11 +94,10 @@ export default class SearchStore {
     if (_.isEmpty(this.phrase)) return;
     const orderBy = { field: 'blockNum', direction: SortBy.DESCENDING };
     let oracles = await searchOracles(this.phrase, null, orderBy);
-    oracles = _.uniqBy(oracles, 'txid');
     oracles = _.uniqBy(oracles, 'topicAddress').map((oracle) => new Oracle(oracle, this.app));
     const topicFilters = [{ status: OracleStatus.WITHDRAW }];
     const topics = await searchTopics(this.phrase, topicFilters);
-    this.withdraws = _.uniqBy(topics, 'txid').map((topic) => new Topic(topic, this.app));
+    this.withdraws = topics.map((topic) => new Topic(topic, this.app));
     this.oracles = _.orderBy(oracles, ['endTime']);
   }
 }
--- a/src/scenes/Search/store.js
+++ b/src/scenes/Search/store.js
@@ -1,6 +1,6 @@
 import _ from 'lodash';
 import { observable, runInAction, action } from 'mobx';
-import { Phases } from 'constants';
+import { Phases, OracleStatus } from 'constants';
 import { searchOracles, searchTopics } from '../../network/graphql/queries';
 import Oracle from '../../stores/models/Oracle';
 import Topic from '../../stores/models/Topic';
@@ -10,7 +10,7 @@ const TAB_SET = 2;
 const TAB_FINALIZE = 3;
 const TAB_WITHDRAW = 4;
 export default class SearchStore {
-  @observable list = [];
+  @observable oracles = [];
   @observable phrase = '';
   @observable loading = false;
   @observable tabIdx = 0;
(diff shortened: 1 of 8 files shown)


BernardHan:
Overall Logic:

backend: 
`searchTopics` now retrieves any topic objects with the specified keywords and in status `WITHDRAW`
`searchOracles` now retrieves any oracle objects with the specified keywords and not in status `WITHDRAW`

frontend:
search/store calls `searchTopics` and `searchOracles`, result from the former is the direct answer to the `withdraw tab` in search result page.
The result from the latter will go through levels of filtering using a `HashMap` to store all `topicAddress`, the filtering goes in order `Finalize` -> `Set Result` -> `Vote` -> `Bet`.
Please refer to `scenes/Search/store.js` line 37 - 40.

src/scenes/Search/store.js
```diff
@@ -30,15 +30,14 @@ export default class SearchStore {
...
       this.loading = false;
       /** the following is a temp work around as searchAll in backend, since we don't have a combined type of Oracle and Topic */
-      this.withdraws = this.list.filter(event => event.phase === Phases.WITHDRAWING && this.topicAddrs.set(event.address, true));
-      this.finalizes = this.list.filter(event => event.phase === Phases.FINALIZING && !this.topicAddrs.has(event.topicAddress) && this.topicAddrs.set(event.topicAddress, true));
-      this.sets = this.list.filter(event => event.phase === Phases.RESULT_SETTING && !this.topicAddrs.has(event.topicAddress) && this.topicAddrs.set(event.topicAddress, true));
-      this.votes = this.list.filter(event => event.phase === Phases.VOTING && event.status === 'VOTING' && !this.topicAddrs.has(event.topicAddress) && this.topicAddrs.set(event.topicAddress, true));
-      this.bets = this.list.filter(event => event.phase === Phases.BETTING && event.status === 'VOTING' && !this.topicAddrs.has(event.topicAddress));
+      this.finalizes = this.oracles.filter(event => event.phase === Phases.FINALIZING && !this.topicAddrs.has(event.topicAddress) && this.topicAddrs.set(event.topicAddress, true));
```
dwalintukan: Another way you could find the latest oracle is to sort it by `blockNum DESC` and get the first item. The higher the `blockNum` the more recent it is..

src/scenes/Search/store.js
```diff
@@ -68,13 +67,14 @@ export default class SearchStore {
...
   async fetch(phrase) {
-    let result = [];
-    if (_.isEmpty(phrase)) return result;
+    if (_.isEmpty(phrase)) return [];
     let oracles = await searchOracles(phrase);
     oracles = _.uniqBy(oracles, 'txid').map((oracle) => new Oracle(oracle, this.app));
-    let topics = await searchTopics(phrase);
+    const topicFilters = [];
+    topicFilters.push({ status: OracleStatus.WITHDRAW });
+    let topics = await searchTopics(phrase, topicFilters);
```
dwalintukan: add the `orderBy` here instead of ordering it after you get the return. you will need to change the `seachTopics` query.
```
export function searchTopics(phrase, filters, orderBy, limit, skip) {
}

const orderBy = { field: 'endTime', direction: SortBy.DESCENDING };
let topics = await searchTopics(phrase, topicFilters, orderBy);
```

src/scenes/Search/store.js
```diff
@@ -68,13 +67,14 @@ export default class SearchStore {
   }
 
   async fetch(phrase) {
-    let result = [];
-    if (_.isEmpty(phrase)) return result;
+    if (_.isEmpty(phrase)) return [];
     let oracles = await searchOracles(phrase);
```
dwalintukan: add the `orderBy` here too instead of doing it after the return.

src/scenes/Search/store.js
```diff
@@ -68,13 +67,14 @@ export default class SearchStore {
...
     oracles = _.uniqBy(oracles, 'txid').map((oracle) => new Oracle(oracle, this.app));
-    let topics = await searchTopics(phrase);
+    const topicFilters = [];
+    topicFilters.push({ status: OracleStatus.WITHDRAW });
```
dwalintukan: inline this with topicFilters: `const topicFilters = [{ status: OracleStatus.WITHDRAW }];`

src/scenes/Search/store.js
BernardHan: I found that the topic object does not have `endTime`, so I think we don't need to sort withdraw events.

src/scenes/Search/store.js
BernardHan: I have changed the logic, plz refer to line 68 and 71, and then 33-36.
Basically I'm using `orderBy` of `blockNum DESC` instead of `endTime`, then I use `lodash uniqBy` of `topicAddress` to remove out-dated duplicates. After that, I sort it by `endTime`. Then I filter out each phase only according to the `event.phase` and `event.status`, getting rid of `HashMap`.

This way is much cleaner, thx for the tip!

FYI, I noticed that our backend supports `orderBy` in `Array`, but our frontend can only handle a single `orderBy` object because of the way how `formatObject` in `graphql/queries.js` parses objects. Do you think it deserves an issue? Although currently a single `orderBy` is enough.

FYI2, our search result page currently is not dynamic as the page won't change if an event transits into another phase. Do you think we should make it dynamic?

BernardHan:
plz note that the searching wait spinner has been added.

src/scenes/Search/store.js
```diff
@@ -64,17 +77,20 @@ export default class SearchStore {
...
-    oracles = _.uniqBy(oracles, 'txid').map((oracle) => new Oracle(oracle, this.app));
-    let topics = await searchTopics(phrase);
-    topics = _.uniqBy(topics, 'txid').map((topic) => new Topic(topic, this.app));
-    result = [...oracles, ...topics];
-    return _.orderBy(result, ['endTime']);
+  async fetch() {
+    if (_.isEmpty(this.phrase)) return;
+    const orderBy = { field: 'blockNum', direction: SortBy.DESCENDING };
+    let oracles = await searchOracles(this.phrase, null, orderBy);
+    oracles = _.uniqBy(oracles, 'txid');
```
dwalintukan: is this necessary to do the `uniqBy`? IIRC, `txid` should be unique among all Oracles.

src/scenes/Search/store.js
```diff
@@ -64,17 +77,20 @@ export default class SearchStore {
...
+    oracles = _.uniqBy(oracles, 'topicAddress').map((oracle) => new Oracle(oracle, this.app));
+    const topicFilters = [{ status: OracleStatus.WITHDRAW }];
+    const topics = await searchTopics(this.phrase, topicFilters);
+    this.withdraws = _.uniqBy(topics, 'txid').map((topic) => new Topic(topic, this.app));
```
dwalintukan: same here, `txid` should be unique among Topics.

src/scenes/Search/styles.js
```diff
@@ -19,6 +19,17 @@ const styles = (theme) => ({
   searchTabContainer: {
     marginTop: theme.sizes.navHeight,
   },
+  row: {
+    display: 'flex',
+    justifyContent: 'center',
+    alignItems: 'center',
+    width: '100%',
+    margin: 20,
```
dwalintukan: use the closest theme values (if it looks ok), `paddingSm`

src/scenes/Search/store.js
BernardHan: I was referring to `scenes/Activities/Finalize`, `scenes/Activities/ResultSetting` and `scenes/Activities/Withdraw`, all these store files do this and I thought it was a safety action.

But yeah, the results show the same if remove it.

src/scenes/Search/store.js
BernardHan: ` syncBlockNum` works like a charm!

I now realize how handy block number is, thx!

src/scenes/Search/store.js
BernardHan: Ok, I will remove these in my test branch

PR #966, opened 2018-09-20
Test module: PendingTxsSnackbarStore

ippudo:
/__mocks__/mixpanelUtil.js is for resolving the conflict between jest and mixpanel.


src/components/PendingTxsSnackbar/store.js
```diff
@@ -1,9 +1,9 @@
 import { observable, action, reaction, runInAction } from 'mobx';
 import _ from 'lodash';
 import { TransactionType, TransactionStatus } from 'constants';
-import { Transaction } from 'models';
 
 import { queryAllTransactions } from '../../network/graphql/queries';
+import { Transaction } from '../../stores/models';
```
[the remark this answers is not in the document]
ippudo: Yes it doesn't work for testing

src/helpers/__mocks__/mixpanelUtil.js
```diff
@@ -0,0 +1,17 @@
+import _ from 'lodash';
+
+let initialized = false;
+
+export default class Tracking {
```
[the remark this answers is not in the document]
ippudo: Some of the store are using Tracking from mixpanel, which will occur error during testing.
The way I can resolve this issue without remove them in stores is to mock it with a function that not using maxpanel.

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,116 @@
...
+      expect(store.count).toBe(2);
+      expect(store.pendingCreateEvents.length).toBe(2);
+      expect(store.isVisible).toBe(true);
+      mockAddTransaction({ txid: 3, status: TransactionStatus.PENDING, type: TransactionType.BET });
+      await store.queryPendingTransactions();
+      expect(store.count).toBe(3);
+      expect(store.pendingCreateEvents.length).toBe(2);
+      expect(store.pendingBets.length).toBe(1);
+      expect(store.isVisible).toBe(true);
+      mockResetTransactionList();
```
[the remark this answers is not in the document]
ippudo: The purpose of mockResetTransactionList() is to remove some transactions from the query result, not to mimic the DB resetting.
I'm not sure whether my knowledge is correct, but I think the total amount of the query results can decreases if some transactions changes their status. e.g. if a transaction change their status from pending to success, it will not appeared at a query for all pending transaction anymore.

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,116 @@
...
+import { mockResetTransactionList, mockAddTransaction } from '../../network/graphql/queries/';
+
+jest.mock('../../network/graphql/queries'); // block and manually mock our backend
+
+describe('PendingTxsSnackbarStore', () => {
+  let store;
+  const app = {
+    global: observable({ syncBlockNum: 0, syncPercent: 10 }),
+  }; // mock the appstore
+  const sleep = ms => new Promise(res => setTimeout(res, ms));
```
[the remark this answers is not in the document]
ippudo: The purpose of this function is to wait for a long period to ensure the reaction is done.
I tried to find a way to test reaction but failed on other way of the async testing.

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+
+  describe('Init', () => {
+    beforeEach(() => {
```
dwalintukan: You can move this entire `beforeEach()` to the root level `describe`. This will run for each test then.

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+      expect(store.pendingFinalizeResults).toEqual(INIT_VALUES.pendingFinalizeResults);
+      expect(store.pendingWithdraws).toEqual(INIT_VALUES.pendingWithdraws);
+      expect(store.pendingTransfers).toEqual(INIT_VALUES.pendingTransfers);
+      expect(store.pendingResetApproves).toEqual(INIT_VALUES.pendingResetApproves);
+      expect.assertions(10);
+    });
+  });
+
+  describe('queryPendingTransactions()', () => {
+    beforeEach(() => {
```
dwalintukan: remove after moving it root level.

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+      await store.queryPendingTransactions();
+      expect(store.count).toBe(1);
+      expect(store.pendingCreateEvents.length).toBe(1);
+      expect(store.isVisible).toBe(true);
+    });
+  });
+
+
+  describe('Reset', () => {
+    beforeEach(() => {
```
dwalintukan: remove after moving to root level.

src/components/PendingTxsSnackbar/store.test.js
dwalintukan: also, shouldn't every test have `init()` called?

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+      expect(store.pendingWithdraws).toEqual(INIT_VALUES.pendingWithdraws);
+      expect(store.pendingTransfers).toEqual(INIT_VALUES.pendingTransfers);
+      expect(store.pendingResetApproves).toEqual(INIT_VALUES.pendingResetApproves);
+      expect.assertions(10);
+    });
+  });
+
+
+  describe('Reactions', () => {
+    beforeEach(async () => {
```
dwalintukan: remove after moving to root level

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+  const sleep = ms => new Promise(res => setTimeout(res, ms));
+
+  describe('Init', () => {
+    beforeEach(() => {
+      store = new PendingTxsSnackbarStore(app); // create a new instance before each test case
+      mockResetTransactionList();
+    });
+
+    it('Init with null tx list', async () => {
+      await store.init();
```
dwalintukan: move this to the root level `beforeEach()` since every test should run init() I believe.

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+
+    it('Fetches all the pending txs and increments the counts', async () => {
+      await store.queryPendingTransactions();
+      expect(store.count).toBe(0);
+      expect(store.isVisible).toBe(false);
+
+      mockAddTransaction({ txid: 1, status: TransactionStatus.PENDING, type: TransactionType.APPROVE_CREATE_EVENT });
+      await store.queryPendingTransactions();
+      expect(store.count).toBe(1);
+      expect(store.pendingBets.length).toBe(0);
```
dwalintukan: although its not a bad practice to check for any side effects, its not necessary to check this. since we know that we are adding an `ApproveCreateEvent tx`, we only really are concerned that`pendingCreateEvents` gets incremented. so checking all the other types is not necessary unless they have already been incremented (from a previous mock add tx). 

think cause and effect.

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+      expect(store.pendingSetResults.length).toBe(0);
+      expect(store.pendingVotes.length).toBe(0);
+      expect(store.pendingFinalizeResults.length).toBe(0);
+      expect(store.pendingWithdraws.length).toBe(0);
+      expect(store.pendingTransfers.length).toBe(0);
+      expect(store.pendingResetApproves.length).toBe(0);
+      expect(store.pendingCreateEvents.length).toBe(1);
+      expect(store.isVisible).toBe(true);
+
+      mockAddTransaction({ txid: 2, status: TransactionStatus.PENDING, type: TransactionType.CREATE_EVENT });
```
dwalintukan: same comment, only necessary to check `count, pendingCreateEvents, and visible`

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+      expect(store.pendingBets.length).toBe(0);
+      expect(store.pendingSetResults.length).toBe(0);
+      expect(store.pendingVotes.length).toBe(0);
+      expect(store.pendingFinalizeResults.length).toBe(0);
+      expect(store.pendingWithdraws.length).toBe(0);
+      expect(store.pendingTransfers.length).toBe(0);
+      expect(store.pendingResetApproves.length).toBe(0);
+      expect(store.isVisible).toBe(true);
+
+      mockAddTransaction({ txid: 3, status: TransactionStatus.PENDING, type: TransactionType.BET });
```
dwalintukan: same comment, only need to check `count, pendingBets, visible`

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+      expect(store.pendingBets.length).toBe(1);
+      expect(store.pendingSetResults.length).toBe(0);
+      expect(store.pendingVotes.length).toBe(0);
+      expect(store.pendingFinalizeResults.length).toBe(0);
+      expect(store.pendingWithdraws.length).toBe(0);
+      expect(store.pendingTransfers.length).toBe(0);
+      expect(store.pendingResetApproves.length).toBe(0);
+      expect(store.isVisible).toBe(true);
+
+      mockAddTransaction({ txid: 4, status: TransactionStatus.PENDING, type: TransactionType.APPROVE_SET_RESULT });
```
dwalintukan: same comment ^

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+      expect(store.pendingBets.length).toBe(1);
+      expect(store.pendingSetResults.length).toBe(1);
+      expect(store.pendingVotes.length).toBe(0);
+      expect(store.pendingFinalizeResults.length).toBe(0);
+      expect(store.pendingWithdraws.length).toBe(0);
+      expect(store.pendingTransfers.length).toBe(0);
+      expect(store.pendingResetApproves.length).toBe(0);
+      expect(store.isVisible).toBe(true);
+
+      mockAddTransaction({ txid: 5, status: TransactionStatus.PENDING, type: TransactionType.SET_RESULT });
```
dwalintukan: same comment ^

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+      expect(store.pendingBets.length).toBe(1);
+      expect(store.pendingSetResults.length).toBe(2);
+      expect(store.pendingVotes.length).toBe(0);
+      expect(store.pendingFinalizeResults.length).toBe(0);
+      expect(store.pendingWithdraws.length).toBe(0);
+      expect(store.pendingTransfers.length).toBe(0);
+      expect(store.pendingResetApproves.length).toBe(0);
+      expect(store.isVisible).toBe(true);
+
+      mockAddTransaction({ txid: 6, status: TransactionStatus.PENDING, type: TransactionType.APPROVE_VOTE });
```
dwalintukan: same comment ^

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+      expect(store.pendingBets.length).toBe(1);
+      expect(store.pendingSetResults.length).toBe(2);
+      expect(store.pendingVotes.length).toBe(1);
+      expect(store.pendingFinalizeResults.length).toBe(0);
+      expect(store.pendingWithdraws.length).toBe(0);
+      expect(store.pendingTransfers.length).toBe(0);
+      expect(store.pendingResetApproves.length).toBe(0);
+      expect(store.isVisible).toBe(true);
+
+      mockAddTransaction({ txid: 7, status: TransactionStatus.PENDING, type: TransactionType.VOTE });
```
dwalintukan: same comment ^

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+      expect(store.pendingBets.length).toBe(1);
+      expect(store.pendingSetResults.length).toBe(2);
+      expect(store.pendingVotes.length).toBe(2);
+      expect(store.pendingFinalizeResults.length).toBe(0);
+      expect(store.pendingWithdraws.length).toBe(0);
+      expect(store.pendingTransfers.length).toBe(0);
+      expect(store.pendingResetApproves.length).toBe(0);
+      expect(store.isVisible).toBe(true);
+
+      mockAddTransaction({ txid: 8, status: TransactionStatus.PENDING, type: TransactionType.FINALIZE_RESULT });
```
dwalintukan: same comment ^

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+      expect(store.pendingBets.length).toBe(1);
+      expect(store.pendingSetResults.length).toBe(2);
+      expect(store.pendingVotes.length).toBe(2);
+      expect(store.pendingFinalizeResults.length).toBe(1);
+      expect(store.pendingWithdraws.length).toBe(0);
+      expect(store.pendingTransfers.length).toBe(0);
+      expect(store.pendingResetApproves.length).toBe(0);
+      expect(store.isVisible).toBe(true);
+
+      mockAddTransaction({ txid: 9, status: TransactionStatus.PENDING, type: TransactionType.WITHDRAW });
```
dwalintukan: same comment ^

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+      expect(store.pendingBets.length).toBe(1);
+      expect(store.pendingSetResults.length).toBe(2);
+      expect(store.pendingVotes.length).toBe(2);
+      expect(store.pendingFinalizeResults.length).toBe(1);
+      expect(store.pendingWithdraws.length).toBe(1);
+      expect(store.pendingTransfers.length).toBe(0);
+      expect(store.pendingResetApproves.length).toBe(0);
+      expect(store.isVisible).toBe(true);
+
+      mockAddTransaction({ txid: 10, status: TransactionStatus.PENDING, type: TransactionType.WITHDRAW_ESCROW });
```
dwalintukan: same comment ^

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+      expect(store.pendingBets.length).toBe(1);
+      expect(store.pendingSetResults.length).toBe(2);
+      expect(store.pendingVotes.length).toBe(2);
+      expect(store.pendingFinalizeResults.length).toBe(1);
+      expect(store.pendingWithdraws.length).toBe(2);
+      expect(store.pendingTransfers.length).toBe(0);
+      expect(store.pendingResetApproves.length).toBe(0);
+      expect(store.isVisible).toBe(true);
+
+      mockAddTransaction({ txid: 11, status: TransactionStatus.PENDING, type: TransactionType.TRANSFER });
```
dwalintukan: same comment ^

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+      expect(store.pendingBets.length).toBe(1);
+      expect(store.pendingSetResults.length).toBe(2);
+      expect(store.pendingVotes.length).toBe(2);
+      expect(store.pendingFinalizeResults.length).toBe(1);
+      expect(store.pendingWithdraws.length).toBe(2);
+      expect(store.pendingTransfers.length).toBe(1);
+      expect(store.pendingResetApproves.length).toBe(0);
+      expect(store.isVisible).toBe(true);
+
+      mockAddTransaction({ txid: 12, status: TransactionStatus.PENDING, type: TransactionType.RESET_APPROVE });
```
dwalintukan: same comment ^

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+      expect(store.pendingBets.length).toBe(1);
+      expect(store.pendingSetResults.length).toBe(2);
+      expect(store.pendingVotes.length).toBe(2);
+      expect(store.pendingFinalizeResults.length).toBe(1);
+      expect(store.pendingWithdraws.length).toBe(2);
+      expect(store.pendingTransfers.length).toBe(1);
+      expect(store.pendingResetApproves.length).toBe(1);
+      expect(store.isVisible).toBe(true);
+
+      mockResetTransactionList();
```
dwalintukan: dont reset (delete) the txs from the list. do what the DB would do, which is change the status of the txs. `PENDING > SUCCESS/FAIL`

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+
+      mockAddTransaction({ txid: 4, status: TransactionStatus.PENDING, type: TransactionType.SET_RESULT });
+      await store.queryPendingTransactions();
+      expect(store.count).toBe(1);
+      expect(store.pendingCreateEvents.length).toBe(0);
+      expect(store.pendingSetResults.length).toBe(1);
+      expect(store.isVisible).toBe(true);
+    });
+
+    it('Fetches all the pending txs and increments the counts', async () => {
```
dwalintukan: duplicate test? this looks like the first test. delete if its testing the same thing.

src/components/PendingTxsSnackbar/store.test.js
```diff
@@ -0,0 +1,317 @@
...
+  });
+
+
+  describe('Reset', () => {
+    beforeEach(() => {
+      store = new PendingTxsSnackbarStore(app); // create a new instance before each test case
+      mockResetTransactionList();
+    });
+
+    it('Run', async () => {
```
dwalintukan: be more descriptive with the test name. `it resets the values`

src/components/PendingTxsSnackbar/store.test.js
dwalintukan: just make `mockResetTransactionList()` to go through each tx and change the status.

src/helpers/__mocks__/mixpanelUtil.js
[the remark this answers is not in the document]
dwalintukan: Let's disable `mixpanelUtil.track()` entirely for non-production env:
```
static track(eventName) {
  if (process.env && process.env.NODE_ENV !== 'production') {
    return;
  }

  ...
}
```

src/components/PendingTxsSnackbar/store.test.js
ippudo: Yes I think you are right. It's too duplicated and many of them can be removed.

But I'm still a little worried about whether we accidentally increased those incorrect params.
Should we add more testing cases like "Fetches a create event txs and check whether the correct param get increased"?

src/components/PendingTxsSnackbar/store.test.js
ippudo: Currently the mocked query function is not using params to filter the result. The returned result are set manually.

I was planning to improve the mock query function later to avoid making huge pr, or do you think I should better do it in this pr?

dwalintukan:
@ippudo you are making the change we discussed with the DB entries' statuses? waiting for that to finish the review.

dwalintukan:
No problem. @ippudo  and @BernardHan should discuss more about the implementation of the testing. It seems like there is code that should be shared between both your testing PRs.

src/network/graphql/__mocks__/dbmodels/mockTransaction.js
```diff
@@ -0,0 +1,54 @@
+import { TransactionStatus } from 'constants';
+
+/*
+* DbModel for Transaction mocks.
+*/
+export default class MockTransaction {
```
dwalintukan: Why do you need a `MockTransaction` model? Is this in reference to the `resetMockTransactionList()` change?

Let me explain this better...

- The `PendingTxSnackbar` queries real live data based on a tx's status, if its `PENDING` it will show up.
- For testing you have a mock DB, and you are creating these txs with PENDING status right?
- So... when you want to emulate the txs being processed and are no longer in `PENDING`, what do you need to do? You just need to change the status of all the txs from `PENDING` -> `SUCCESS/FAIL`.
- You should be creating a helper function in `mockData.js`. This is your mock DB server right? It does the "insert" into the DB, so therefore you should have a function that does "update" to it as well.
- Not sure the whole logic of creating this MockTransaction class... Move your `setSuccess()` and `setFail()` helper functions to `mockData.js` and have your `resetMockTransactionList()` call this for every tx in your mock DB.

ippudo: @dwalintukan 
Hi deric thank you for review my pr. I think this could be a better way for testing.
These models designed to have the same behavior with the object that a graphql request should return, but can be with more helpful feature, like: 
* data validation (to avoid making wrong callback data and wasting time because of these typo)
* auto filling (fill default data to blank field to make it more like a real callback data, also less copy & paste)
* command chain design pattern (In future we may also have to test topic and oracle, and when mocking topic/oracle we may also need to change related oracles/txs data. if we just use manually generated object for them, we also need to maintain all of these changes one by one, but with command chain we can easily connect related topic/oracle/txs together, then we change one, the other can also be updated automatically )
For only transaction it surely not necessary, but I think it may become helpful when we start testing oracles and topics. I'm glad to know your opinion on this ideas.

ippudo: And currently this model is used like this:

tx1 = mockAddTransaction(1, { type: TransactionType.APPROVE_CREATE_EVENT });
theFunctionWeWantToTestWithPending();
// when we want to set it to success
tx1.setSuccess();
theFunctionWeWantToTestWithSuccess();

dwalintukan: @ippudo 
I have no issue with having these helper methods that are useful for testing (setSuccess, setFail, etc).

But, the reason I do not want to have `MockTransaction` classes is for the reason of overhead of future changes. If we were using Typescript and we had inheritance, I would say definitely do the `MockTransaction` classes by extending the `Transaction` class. But we are using JS so there is no inheritance.

So say we do have the `Transaction` and `MockTransaction` class. I can "almost" guarantee later down the road that if we have to change the `Transaction` class (say adding a new field), that someone will miss it adding it to the `MockTransaction` class. I have seen this countless times in my career. So, adding the `MockTransaction` class adds this extra overhead where we have to change the class in two places vs one and gives another vector of possibly introducing a bug.

That being said, I would like to keep the testing on the original models. And when you need something specific for testing, create a helper function, e.g:
```
function mockAddTransaction(args) {
  const tx = new Transaction({
    type = '' // One of: TransactionType
    txid = '' // Transaction ID assigned by the blockchain
    status = '' // One of: TransactionStatus
    createdTime = '' // UNIX timestamp when Transaction was created
    // etc, etc, set your default values here
  });
  Object.assign(tx, args); // overwrite any default values with the ones passed in
}

function mockSetTxStatus(tx, status) {
  tx.status = status;
}

function setAllTxsSuccess(txs) {
  _.each(txs, tx => tx.status = 'SUCCESS';
}
```

dwalintukan: @ippudo 
We are using the UI Transaction class throughout the entire UI. I understand that you are trying to match the backend structure, but every time we call `queryAllTransactions` in the UI we are always mapping it to the UI Transaction model: `const txs = map(result, (tx) => new Transaction(tx));`. Don't think there are any cases where we are using the raw Transaction callback data without mapping it to the Transaction class.

Even in `PendingTxsSnackbarStore`, which is the subject of your tests, you are testing against the UI Transaction model. If you want to validate the backend data that should be validated on the server-side tests. In this case, we are validating the UI front end Transactions.

ippudo: @dwalintukan 
Maybe that is where the problem comes. We currently map the result to the UI model after each `queryAllTransactions()` called, not in `queryAllTransactions()`. And the function I'm mocking is also `queryAllTransactions()`, so if I use UI TX models as callback data, we will use these UI TX models to generate another UI TX model.

src/components/PendingTxsSnackbar/store.test.js
dwalintukan: remove, already called in beforeEach

src/stores/AllEventsStore.test.js
```diff
@@ -11,30 +15,46 @@ const localStorageMock = {
...
     refreshing: false,
     ui: { location: 0 },
-  }; // mock the appstore
+  }); // mock the appstore
   beforeEach(() => {
     store = new AllEventsStore(app); // create a new instance before each test case
+    mockResetOracleList();
+    mockResetTopicList();
+    for (let i = 0; i < 18; i++) {
+      mockAddTopic({ txid: i });
```
dwalintukan: use `cryptoRandomString(64)` for all txids

src/stores/AllEventsStore.test.js
```diff
@@ -11,30 +15,46 @@ const localStorageMock = {
...
+      mockAddOracle({
+        txid: i,
```
dwalintukan: use cryptoRandomString(64) for all txids

ippudo:
More tests on the way!

PR #971, opened 2018-09-22
Bookmark Feature

ippudo:
These code come from a branch 1 month ago and I'm feeling no reason to dispose it.


src/scenes/QtumPrediction/components/EventCard/index.js
```diff
@@ -88,7 +89,7 @@ export default class EventCard extends PureComponent {
               {isPending && <EventWarning id="str.pendingConfirmation" message="Pending Confirmation" />}
               {isUpcoming && <EventWarning id="str.upcoming" message="Upcoming" type="upcoming" />}
               <Typography variant="headline" className={classes.eventCardName}>
-                {name}
+                {name}<FavoriteButton event={this.props.event} />
```
dwalintukan: same comment as the other one. suggest to create a separate component for the Event name + fav button.

src/components/EventCard/index.js
```diff
@@ -90,7 +91,7 @@ export default class EventCard extends PureComponent {
               {isPending && <EventWarning id="str.pendingConfirmation" message="Pending Confirmation" />}
               {isUpcoming && <EventWarning id="str.upcoming" message="Upcoming" type="upcoming" />}
               <Typography variant="headline" className={classes.eventCardName}>
-                {name}
+                {name}<FavoriteButton event={this.props.event} />
```
dwalintukan: move `FavoriteButton` out of `Typography` and wrap both in a `span` or `div`. yes it does work this way, but its not proper since its not meant to be part of the Typography.

src/components/EventCard/FavoriteButton.js
```diff
@@ -0,0 +1,45 @@
+import React, { Component, Fragment } from 'react';
+import { inject, observer } from 'mobx-react';
+import { withRouter } from 'react-router';
+import {
```
dwalintukan: nit: move to the same line. I disabled this rule for eslint so it doesn't apply the max-len rule for imports.

src/scenes/Activities/Favorite/store.js
```diff
@@ -0,0 +1,92 @@
+import { observable, action, runInAction, reaction } from 'mobx';
+import _ from 'lodash';
```
dwalintukan: import only the necessary modules needed (using destructuring)

src/scenes/Activities/Favorite/store.js
```diff
@@ -0,0 +1,92 @@
...
+import Topic from '../../../stores/models/Topic';
+import Oracle from '../../../stores/models/Oracle';
+
+const INIT_VALUES_FAVPAGE = {
+  loaded: false, // loading state?
+  loadingMore: true, // for loading icon?
+  displayList: [], // data list
+};
+
+export default class {
```
dwalintukan: nit: add the class name. I find its easier to tell what class you are looking at.

src/scenes/Activities/Favorite/store.js
```diff
@@ -0,0 +1,92 @@
...
+      () => this.app.wallet.addresses + this.app.global.syncBlockNum,
+      () => {
+        if (this.app.ui.location === Routes.FAVORITE) {
+          this.init();
+        }
+      }
+    );
+    reaction(
+      () => this.displayList,
+      () => {
```
dwalintukan: empty reaction?

src/scenes/Activities/Favorite/store.js
```diff
@@ -0,0 +1,92 @@
...
+    this.app.ui.location = Routes.FAVORITE; // change ui location, for tabs to render correctly
+    this.displayList = await this.fetchFav();
+  }
+
+  fetchFav = async () => {
+    if (this.favList.length === 0) return [];
+    // Get all event in WITHDRAW phase as Topic at favorite topic address list "favList"
+    const topicOrderBy = { field: 'endTime', direction: SortBy.ASCENDING };
+    const topicFilters = this.favList.map(topicAddress => ({ address: topicAddress, status: OracleStatus.WITHDRAW }));
+    const topics = await queryAllTopics(topicFilters, topicOrderBy, Infinity, Infinity);
```
dwalintukan: remove `skip` (the last param) since we dont want to skip any. and change `limit` to a big number like 5000. the default limit should be 50 so we need to increase it.

`Infinity` is not actually being used, `queryAllTopics` checks if those values are finite.

src/scenes/Activities/Favorite/store.js
```diff
@@ -0,0 +1,92 @@
...
+    const topicResult = _.uniqBy(topics, 'txid').map((topic) => new Topic(topic, this.app));
```
dwalintukan: `uniqBy` is not necessary. There should only be 1 txid per topic ever. Meaning you shouldn't ever get duplicates.

src/scenes/Activities/Favorite/store.js
```diff
@@ -0,0 +1,92 @@
...
+    );
+  }
+
+  updateLocalStorageFavList() {
+    localStorage.setItem('bodhi_dapp_favList', JSON.stringify(this.favList));
+  }
+
+  @action
+  setFavorite = (topicAddress) => {
+    if (this.isInFavorite(topicAddress)) this.favList = this.favList.filter(x => x !== topicAddress);
```
dwalintukan: nit: move the filter logic to next line in brackets since we have an else statement

src/scenes/Activities/Favorite/store.js
```diff
@@ -0,0 +1,92 @@
...
+    else {
+      this.favList.push(topicAddress);
+      this.favList.replace(this.favList);
```
dwalintukan: why do you need replace? `this.favList.push(topicAddress)` should handle any refreshes needed since its observable.

src/scenes/Activities/Favorite/store.js
```diff
@@ -0,0 +1,92 @@
...
+    }
+    this.updateLocalStorageFavList();
+  }
+
+  @action
+  isInFavorite = (topicAddress) => this.favList.some(x => x === topicAddress)
```
dwalintukan: prefer to use `find` or `findIndex`

src/scenes/Activities/Favorite/store.js
```diff
@@ -0,0 +1,92 @@
...
+    const topicOrderBy = { field: 'endTime', direction: SortBy.ASCENDING };
+    const topicFilters = this.favList.map(topicAddress => ({ address: topicAddress, status: OracleStatus.WITHDRAW }));
+    const topics = await queryAllTopics(topicFilters, topicOrderBy, Infinity, Infinity);
+    const topicResult = _.uniqBy(topics, 'txid').map((topic) => new Topic(topic, this.app));
+
+    // For those events which is not in WITHDRAW phase, search for the latest oracle
+    const locatedTopics = topicResult.map(topicObject => (topicObject.address));
+    const oracleFilters = _.difference(this.favList, locatedTopics).map(omittedTopicAddress => ({ topicAddress: omittedTopicAddress }));
+    const oracleOrderBy = { field: 'blockNum', direction: SortBy.DESCENDING };
+    const oracles = await queryAllOracles(oracleFilters, oracleOrderBy, Infinity, Infinity);
```
dwalintukan: remove `skip` and set `limit` to 5000

src/scenes/Activities/Favorite/store.js
```diff
@@ -0,0 +1,92 @@
...
+    const oracleResult = _.uniqBy(oracles, 'topicAddress').map((oracle) => new Oracle(oracle, this.app));
```
dwalintukan: shouldnt need `uniqBy`

src/scenes/Activities/Favorite/index.js
```diff
@@ -0,0 +1,32 @@
...
+    this.props.store.favorite.init();
+  }
+
+  render() {
+    const { displayList, loadingMore, loading } = this.props.store.favorite;
+    if (loading) return <Loading />;
+    const events = (displayList || []).map((event, i) => <EventCard key={i} index={i} event={event} />); // eslint-disable-line
+    return (
+      <Fragment>
+        <InfiniteScroll
```
dwalintukan: dont use `InfiniteScroll` if its not using the loadMore functionality

src/components/EventCard/FavoriteButton.js
```diff
@@ -0,0 +1,45 @@
...
+
+import styles from './styles';
+
+@withStyles(styles, { withTheme: true })
+@withRouter
+@inject('store')
+@observer
+class FavoriteButton extends Component {
+  state = {
+    forceRefreshFlag: false,
```
dwalintukan: why do you need this? we should be able to handle refreshes based on the observing changes in the store.

src/scenes/Activities/Favorite/store.js
ippudo: OracleResult will return all phases of oracles. uniqBy here will remove unnecessary phases. A similar logic is also used in search store to keep only the latest phase.

src/components/EventCard/FavoriteButton.js
dwalintukan: Yes let's avoid hacky solutions to make things work. Should be able to get this to work using Mobx and observable values.

src/components/EventCard/styles.js
```diff
@@ -37,7 +37,7 @@ const styles = (theme) => ({
   eventCardName: {
     marginBottom: theme.padding.xs.px,
     display: '-webkit-box',
-    maxWidth: '400px',
+    width: '400px',
```
dwalintukan: try adding `flex: 1` to this, and make sure the parent container has `display: flex`. read about using flexbox. flex: 1 will take up all the space of the parent container.

this solution of setting a width is not good for responsive design.

src/scenes/QtumPrediction/components/EventCard/styles.js
```diff
@@ -37,7 +37,7 @@ const styles = (theme) => ({
   eventCardName: {
     marginBottom: theme.padding.xs.px,
     display: '-webkit-box',
-    maxWidth: '700px',
+    width: '700px',
```
dwalintukan: same thing, use flexbox to take up all the space. avoid setting static width/heights.

src/scenes/Activities/Favorite/store.js
```diff
@@ -0,0 +1,81 @@
...
+    if (this.isInFavorite(topicAddress)) {
+      this.favList = this.favList.filter(x => x !== topicAddress);
+    } else {
+      this.favList.push(topicAddress);
+    }
+    this.updateLocalStorageFavList();
+  }
+
+  @action
+  isInFavorite = (topicAddress) => this.favList.find(x => x === topicAddress)
```
dwalintukan: you will have to use the `not not operator` to turn it to a boolean: 
`!!this.favList.find(x => x === topicAddress)`

believe this currently is returning the object itself, but looks like you want a boolean.

src/scenes/Activities/Favorite/index.js
dwalintukan: Yes agreed. We should have an `EventCardGrid` component and the `InfiniteScroll` can use `EventCardGrid` within it as well.

Issue #923, opened 2018-09-05
Show error message on bad API request


dwalintukan:
We are onlying show the status code and not the error message for a bad request. We should be showing the error message like the one we can see in the console: `error: Address does not exist`.

![image](https://user-images.githubusercontent.com/4350404/45086278-f90f3180-b12c-11e8-8a24-27e19a353322.png)

![image](https://user-images.githubusercontent.com/4350404/45086308-06c4b700-b12d-11e8-993c-1d60868b31d9.png)

PR #979, opened 2018-09-25
#923 Error Message

ippudo:
Much easier than estimated


dwalintukan:
I was actually referring to do this for all error messages, but can handle this in a diff PR.

ippudo:
It's my fault. Will do it tomorrow with a new pr.

dwalintukan:
you can just use this PR and add to it.

PR #982, opened 2018-09-26
Qrypto message fixes

dwalintukan:


src/stores/models/WalletAddress.js
```diff
@@ -1,9 +1,11 @@
+import { observable } from 'mobx';
+
 import { satoshiToDecimal } from '../../helpers/utility';
 
 export default class WalletAddress {
-  address = ''
-  qtum = 0
-  bot = 0
+  @observable address = '';
```
BernardHan: semicolon or not?

src/stores/GlobalStore.js
```diff
@@ -74,33 +74,53 @@ export default class GlobalStore {
...
+   * @param {MessageEvent} event Message to handle.
+   */
+  handleWindowMessage = (event) => {
+    if (event.data.message && event.data.message.type) {
+      const types = {
+        QRYPTO_INSTALLED_OR_UPDATED: this.handleQryptoInstall,
+        ACCOUNT_CHANGED: this.handleQryptoAccountChange,
+      };
+      const messageAction = types[event.data.message.type];
+      if (messageAction) messageAction(event);
```
BernardHan: what does this do?

src/stores/models/WalletAddress.js
dwalintukan: it seems to be ok without the semi colon. force of habit to add it. I prefer it there. what about you?

src/stores/GlobalStore.js
dwalintukan: This checks the `window message event` and if any of them are `QRYPTO_INSTALLED_OR_UPDATED` or `ACCOUNT_CHANGED` it will execute their functions and pass the `event` param along with it.

src/stores/models/WalletAddress.js
dwalintukan: sounds good. ya i like semicolons too 👍

src/stores/GlobalStore.js
dwalintukan:
```
// so you have your types. these correspond to a function in the class:
const types = {
        QRYPTO_INSTALLED_OR_UPDATED: this.handleQryptoInstall,
        ACCOUNT_CHANGED: this.handleQryptoAccountChange,
      };

// this fetches the proper function to execute, e.g. types[QRYPTO_INSTALLED_OR_UPDATED] = this.handleQryptoInstall. messageAction is now pointing the the function.
const messageAction = types[event.data.message.type];

// if messageAction is a defined function, execute the function: messageAction(event) === this.handleQryptoInstall(event)
if (messageAction) messageAction(event);
```

PR #991, opened 2018-09-27
ButtonGlobalOverrides

GoodThomas:


src/config/theme.js
```diff
@@ -297,4 +270,41 @@ export const theme = {
 
 const bodhiTheme = createMuiTheme(theme);
 
+bodhiTheme.overrides = {
+  ...bodhiTheme.overrides,
+  MuiButton: {
```
dwalintukan: move this back into `theme.overrides`. do it inline.

src/config/theme.js
```diff
@@ -297,4 +270,41 @@ export const theme = {
...
+    root: {
+      borderRadius: px(paddingLg),
+      textTransform: 'none',
+      [bodhiTheme.breakpoints.down('xs')]: {
+        padding: '7px 8px',
```
dwalintukan: lets use `paddingXs` all around.

src/config/theme.js
GoodThomas: I want to use 'breakpoints' in 'overrides', but I didn't find a better way.

dwalintukan: got it :+1:

PR #1004, opened 2018-10-01
989 998 tx cost error & explorer link fixes

### LINKED ISSUE #989: Fix txid click to go to proper explorer

dwalintukan:
#989 
#998


src/scenes/CreateEvent/store.js
```diff
@@ -252,8 +252,20 @@ export default class CreateEventStore {
       return;
     }
 
-    if (await this.hasPendingCreateTxs()) return;
-    if (await !this.getEscrowAmount()) return;
+    // Close if getting pending txs fails
+    const hasPendingTxs = await this.hasPendingCreateTxs();
```
BernardHan: OK with this but curious why not

```js
if (await this.hasPendingCreateTxs()) {
    this.close();
    return;
}
```

src/scenes/CreateEvent/store.js
```diff
@@ -325,6 +339,7 @@ export default class CreateEventStore {
   /**
    * Gets the average block time from the Insight API.
    */
+  @action
```
BernardHan: averageBlockTime is not set to @observable

src/scenes/CreateEvent/store.js
```diff
@@ -310,6 +323,7 @@ export default class CreateEventStore {
    * Fetches the escrow amount from the API.
    * @return {boolean} True if the API call was successful.
    */
+  @action
```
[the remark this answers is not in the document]
BernardHan: But it works, how?

src/scenes/CreateEvent/store.js
```diff
@@ -252,8 +252,20 @@ export default class CreateEventStore {
...
-    if (await !this.getEscrowAmount()) return;
+    // Close if getting pending txs fails
+    const hasPendingTxs = await this.hasPendingCreateTxs();
+    if (hasPendingTxs) {
+      this.close();
+      return;
+    }
+
+    // Close if unable to get the escrow amount
+    const escrowAmountSuccess = await this.getEscrowAmount();
```
BernardHan: By playing around, I notice that
```js
if (!(await this.getEscrowAmount())) return;
```
will achieve the same. Previously we had `!` before `this.getEscrowAmount`, which did not work.

Why it did not work? I'm guessing it's because `await` was waiting for `!` not `this.getEscrowAmount`. But I'm doubting, do you know the exact reason?

src/scenes/CreateEvent/store.js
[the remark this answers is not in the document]
dwalintukan: oh good catch. changed to observable.

src/scenes/CreateEvent/store.js
dwalintukan: Oh I think you are misunderstanding the `action` keyword:
https://mobx.js.org/refguide/action.html#when-to-use-actions

I mark action whenever the state changes.

src/scenes/CreateEvent/store.js
dwalintukan: Ah ya you are right, adding the not operator with the await enclosed in brackets will do the same `(!(await this.getEscrowAmount()))`.

This does not work:
`if (await !this.getEscrowAmount()) return;`
because `this.getEscrowAmount()` returns a Promise so `awaiting` it returns the proper `boolean` return.

BernardHan: oh ok, but what does `!` catch if `this.getEscrowAmount()` not finished? Does it just catch `true` since it did not reach `return;`?

src/scenes/CreateEvent/store.js
BernardHan: ah I thought you meant those to be observable, alright.

src/scenes/CreateEvent/store.js
BernardHan: You don't need to change if you think both are equal.

Issue #990, opened 2018-09-27
Clean up bad pending txs


dwalintukan:
- Add reset approve logic to result setting oracle detail
- Checks for loggedIn status of qrypto before prompting follow up tx

PR #1005, opened 2018-10-01
990 pending txs cleanup

dwalintukan:
#990 
- Add reset approve logic to result setting oracle detail
- Checks for loggedIn status of qrypto before prompting follow up tx
- Add `topicAddress` and `oracleAddress` to reset approve mutation to track it on detail pages
- Replace Badge component for My Activities with custom comp
- Style settings page

image
image
image


BernardHan:
I'm getting console error in this PR:
image

Timing when I get this error:
- everytime when I first open the website, I get exactly one, but no error when I go around `Qtum Prediction`, `BOT Court` and `All Event`.
- sometimes when I click a tab in `My Activities`, `Setting`, I get an error

src/stores/TransactionStore.js
```diff
@@ -249,15 +264,20 @@ export default class TransactionStore {
...
    */
   @action
-  addResetApproveTx = async (spender) => {
+  addResetApproveTx = async (spender, topicAddress, oracleAddress) => {
     this.transactions.push(observable.object(new Transaction({
       type: TransactionType.RESET_APPROVE,
       senderAddress: this.app.wallet.currentAddress,
       receiverAddress: spender,
+      topicAddress,
+      oracleAddress,
```
BernardHan: where are those two used? I don't get what do you mean by "track on detail page"

src/scenes/Event/store.js
```diff
@@ -555,34 +562,26 @@ export default class EventStore {
...
     const { topicAddress } = this.oracle;
     const oracleAddress = this.oracle.address;
     const optionIdx = this.selectedOption.idx;
     const amountSatoshi = decimalToSatoshi(this.amount);
-    const allowance = await checkAllowance(currentAddress, topicAddress);
 
-    if (isAllowanceEnough(allowance, amountSatoshi)) {
-      await this.app.tx.addSetResultTx(undefined, topicAddress, oracleAddress, optionIdx, amountSatoshi);
-    } else {
+    if (this.allowance > 0 && !isAllowanceEnough(this.allowance, amountSatoshi)) {
```
BernardHan: Previously we did not reset for setResult, but it worked fine. And so I'm a little confused, what was the reason to implement this reset?

My mind tells me it's a preparation for ERC20?
So even though QRC does not need to reset, we still let it behave like ERC20, am I wrong?

dwalintukan: FYI, this is a super edge case that shouldn't normally happen for most people, but it did for me while I was testing a lot.

On one of the events in the Set Result phase, I had previously had set a bad approve amount due to testing. Set Result is supposed to be 100 BOT, or 10000000000 Botoshi. While implementing a feature/testing I approved for 100 Botoshi on this event (this won't happen to anyone now since the code works).

But, that being said, without the RESET_APPROVE call, my approval for this event would be stuck at 100 Botoshi. So I would never be able to set the result on this event unless I reset the approve back to 0 first, then approve for the correct 10000000000 Botoshi.

Granted this was for a super edge case that should only happen for developers, smart contracts are open for anyone to interact with it. So say if a power user (person with technical Solidity knowledge) decided to call approve for this event outside of our dapp, when they used our dapp they would be stuck without a way to ever set the result on this event. I know this is prob a case that will never happen, but I did it more for myself to be able to reset the approve on the events that I already set prior wrong approve amounts for.

dwalintukan:
@BernardHan are you using the latest qrypto 1.4.2? Just placed a bet and it was fine.

dwalintukan:
@BernardHan I know what happened. Your Qrypto session timed out (the default is 10 mins. can be changed in settings). Going to make changes to handle this case.

dwalintukan:
I added a status indicator for the wallet:
image

Also, the ExecuteTxDialog disappears if the user is logged out so it should fix that issue you were having.

src/stores/TransactionStore.js
BernardHan: Oh so we want to show reset histories to users

src/scenes/Event/store.js
BernardHan: wow, thank you

PR #1012, opened 2018-10-03
Combine 2 tables into one table with tab on Event Pages

ippudo:
image

Initial design data from Lili.

We do this because in the latest designed event page we will have more tables in recent future.


src/scenes/Event/components/HistoryTable/index.js
```diff
@@ -0,0 +1,71 @@
...
+const { UNCONFIRMED, TOPIC, ORACLE } = EventType; // eslint-disable-line
+
+@withStyles(styles, { withTheme: true })
+@injectIntl
+@inject('store')
+@observer
+export default class HistoryTable extends Component {
+  static propTypes = {
+    classes: PropTypes.object.isRequired,
+    // oracles: PropTypes.array.isRequired,
```
dwalintukan: remove

src/scenes/Event/components/HistoryTable/styles.js
```diff
@@ -0,0 +1,18 @@
...
+  pickerTab: {
+    margin: 0,
+    marginTop: 0,
+    paddingTop: 10,
+    padding: 10,
+    flexGrow: 1,
+  },
+  subTableContainer: {
+    // height: 400,
+    // overflowY: 'auto',
```
dwalintukan: remove unused

src/scenes/Event/components/HistoryTable/styles.js
```diff
@@ -0,0 +1,18 @@
+const styles = (theme) => ({
+  pickerTab: {
+    margin: 0,
+    marginTop: 0,
+    paddingTop: 10,
+    padding: 10,
```
dwalintukan: remove all margin and padding. we already define `MuiTab` in global theme overrides so lets use the default version to handle all the margins.

src/languageProvider/locales/en_US.json
```diff
@@ -224,7 +224,7 @@
   "str.time": "Time",
   "str.to": "To",
   "str.token": "Token",
-  "str.transaction": "TRANSACTIONS",
+  "str.transaction": "My Transactions",
```
dwalintukan: remove this string and make a new string called `str.myTransactions`. We already have existing translations in Korean and Chinese for `str.transaction` so if any part of the string changes in English we should be removing any old translations and update with new translations.

src/config/theme.js
```diff
@@ -223,6 +223,7 @@ export const theme = {
       root: {
         marginTop: px(paddingUnit),
         marginBottom: px(paddingUnit),
+        borderBottom: `1px solid ${textColorLight}`,
```
dwalintukan: use `theme.palette.background.grey` which is our border color.

src/scenes/Event/components/HistoryTable/index.js
```diff
@@ -0,0 +1,71 @@
...
+    intl: intlShape.isRequired, // eslint-disable-line react/no-typos
+  };
+  state = {
+    tabValue: 'transactionHistory',
```
dwalintukan: lets create a `const TAB_TYPE = ['resultHistory', 'transactionHistory']` to handle the tab values.

src/scenes/Event/components/HistoryTable/index.js
```diff
@@ -0,0 +1,71 @@
...
+import { inject, observer } from 'mobx-react';
+import PropTypes from 'prop-types';
+import { injectIntl, intlShape } from 'react-intl';
+import { EventType } from 'constants';
+import { Tabs, Tab, withStyles } from '@material-ui/core';
+
+import styles from './styles';
+import { ResultHistory, TransactionHistory } from '../';
+
+const { UNCONFIRMED, TOPIC, ORACLE } = EventType; // eslint-disable-line
```
dwalintukan: why do you need to disable eslint?

src/scenes/Event/components/HistoryTable/index.js
```diff
@@ -0,0 +1,71 @@
...
+    classes: PropTypes.object.isRequired,
+    // oracles: PropTypes.array.isRequired,
+    intl: intlShape.isRequired, // eslint-disable-line react/no-typos
+  };
+  state = {
+    tabValue: 'transactionHistory',
+  };
+
+  componentDidMount = () => {
+    this.setState({ tabValue: this.props.resultHistory ? 'resultHistory' : 'transactionHistory' });
```
dwalintukan: should be able to set this directly in `state = {}`?

dwalintukan: ok np

Issue #975, opened 2018-09-25
Add loading state for Create Event


dwalintukan:

dwalintukan:
@BernardHan Please note I changed the order of the logic in my `819-820-qrypto` branch. Best to branch off this one or wait til it gets in master.

PR #1016, opened 2018-10-04
Add Loading to Create Event

BernardHan:
#975 

I need different `marginTop` for the loading spinner, so I used `cx` from `classnames` to conditionally choose marginTop in styles.js.

Not sure if this is good enough, any suggestion?

image

### DIFF
--- a/src/components/InfiniteScroll/index.js
+++ b/src/components/InfiniteScroll/index.js
@@ -70,7 +70,7 @@ export default class InfiniteScroll extends Component {
   }
 }
 
-const LoadingMore = () => <Row><Loading text={messages.loadMoreMsg} event /></Row>;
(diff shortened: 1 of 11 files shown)


src/components/InfiniteScroll/index.js
```diff
@@ -70,7 +70,7 @@ export default class InfiniteScroll extends Component {
   }
 }
 
-const LoadingMore = () => <Row><Loading text={messages.loadMoreMsg} /></Row>;
+const LoadingMore = () => <Row><Loading text={messages.loadMoreMsg} event /></Row>;
```
henryzbw: where is this event from?

dwalintukan:
Ditto what Henry said ^. Maybe a bit more padding.

Issue #1022, opened 2018-10-04
Activity history remains to be previous wallet's after switch to another wallet in qrypto, until the page gets refreshed.


ippudo:
We needs a refresh for it. 

BTW I'm not sure whether activity history is still designed to show only the current address' activity?

dwalintukan:
My Activites > Activity History should only show your wallet's txs. Transaction History in event detail should show all txs by everyone.

Issue #1021, opened 2018-10-04
When switch to a new wallet, transactions cached for ExecuteTxDialog remains to be previous one's


ippudo:
And you still can send these transactions, then it will appeared as the activity history of previous wallet's.

ippudo:
We still need to handle the following flow:

Wallet A -> Bet -> close Transaction confirm dialog -> change to Wallet B -> the Bet transaction from A still at the confirm dialog

dwalintukan:
#1034

Issue #1019, opened 2018-10-04
Current block time and block number does not show up


dwalintukan:
As reported by Jack:
![image](https://user-images.githubusercontent.com/4350404/46451887-09b1d680-c7c3-11e8-962d-d8b6c196bd8a.png)

He says sometimes it disappears when doing actions on the site.

PR #1024, opened 2018-10-04
1021 follow up addr

dwalintukan:
#1019
#1021 
#1022


dwalintukan:
@BernardHan From the EventPage and CreateEvent stores, I am passing `currentWalletAddress` because that is the old flow (using the Qtum local wallet). 

In TransactionStore, I assign it from `tx.senderAddress`.

The reason for this is because when using Qrypto (non-local wallet) the user can change addresses so say if a user does an `ApproveVote` with address A. When the approve succeeds with address A, but the user switched to address B, the follow up `Vote` tx that is added should have the same address A. Then we verify what the `Vote` tx is added if the `tx.senderAddress === currentWalletAddress`.

Issue #1030, opened 2018-10-04
Activity History Pagination Doesn't Work


ippudo:
Try to click buttons at the bottom of the table.

PR #1035, opened 2018-10-05
Activity history fixes

### LINKED ISSUE #1032: activities history page won't refresh when new block comes

dwalintukan:
#1032 
#1030


ippudo:
I'm a little bit worried about that it looks like we transferred a lot of light weight Pure Component to Component links to the store. Will it affect the efficiency for the whole app?

PR #1039, opened 2018-10-06
return mutated object to avoid on query

henryzbw:
Only made change for bet now
want to make sure it's the way you want
and then continue to others

don't need to change server now

### DIFF
--- a/src/network/graphql/mutations.js
+++ b/src/network/graphql/mutations.js
@@ -69,5 +69,6 @@ class GraphMutation {
  */
 export async function createTransaction(mutationName, args) {
   const res = await new GraphMutation(mutationName, args, TYPE.transaction).execute();
-  return new Transaction(res);
+  const { data: { [mutationName]: ret } } = res;
+  return new Transaction(ret);
 }
--- a/src/network/graphql/mutations.js
+++ b/src/network/graphql/mutations.js
@@ -1,5 +1,6 @@
 import gql from 'graphql-tag';
 import { each, isFinite } from 'lodash';
+import { Transaction } from 'models';
 
 import client from './';
 import { TYPE, getMutation, isValidEnum } from './schema';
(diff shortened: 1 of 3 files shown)


src/stores/TransactionStore.js
```diff
@@ -558,7 +558,7 @@ export default class TransactionStore {
           amount,
         });
 
-        await this.onTxExecuted(index, tx);
+        await this.onTxExecuted(index, tx, new Transaction(createBet));
```
dwalintukan: let's make a change to `mutations.js` where `createTransaction()` returns a `new Transaction()` instance.

henryzbw:
@dwalintukan verified it works. please have a look at the code after your last review. If it's okay, i will ship it.

src/stores/TransactionStore.js
```diff
@@ -954,13 +954,12 @@ export default class TransactionStore {
   @action
   executeTransfer = async (index, tx) => {
     try {
-      const { data: { transfer } } = await createTransaction('transfer', {
+      const newTx = await createTransaction('transfer', {
```
dwalintukan: nit: lol why is this one named different than the other ones?

src/network/graphql/mutations.js
```diff
@@ -66,6 +67,8 @@ class GraphMutation {
  * @param {string} mutationName Name of the mutation.
  * @param {object} args Arguments for the mutation.
  */
-export function createTransaction(mutationName, args) {
-  return new GraphMutation(mutationName, args, TYPE.transaction).execute();
+export async function createTransaction(mutationName, args) {
+  const res = await new GraphMutation(mutationName, args, TYPE.transaction).execute();
+  const { data: { [mutationName]: ret } } = res;
```
dwalintukan: nit: `[mutationName]: tx`

PR #1041, opened 2018-10-07
Translations

dwalintukan:


src/languageProvider/locales/zh-Hans.json
```diff
@@ -374,7 +375,17 @@
...
+  "txConfirm.txOne": "交易确认 1/2。当前交易确认成功后，您将需要确认另外一个交易。",
+  "txConfirm.txTwo": "交易确认 2/2。",
+  "txType.approve": "批准",
+  "txType.bet": "投注",
+  "txType.createEvent": "创建事件",
+  "txType.finalizeResult": "结束",
+  "txType.setResult": "设置结果",
+  "txType.vote": "仲裁",
+  "txType.withdraw": "提现",
+  "txType.withdrawEscrow": "撤回托管",
```
ippudo: There's an inconsistency here: "str.withdrawEscrow": "提现押金"

What does this transaction type for? Is it for withdraw the 5 BOT that used when create an event? Then it should be "提现押金".

dwalintukan: Yes It is for withdrawing the escrow that is needed when creating an event.

dwalintukan: changed. thanks for the updated translation.

Issue #1036, opened 2018-10-05
Remove scrollbar and increase Create Event dialog width


dwalintukan:
The width is smaller than it was before. Looks scrunched up. Make it wider like it was before.

![image](https://user-images.githubusercontent.com/4350404/46523598-66d78600-c8b0-11e8-938a-f4b6880aee11.png)

dwalintukan:
think we can just remove the scrollbar entirely? might look better without it.

BernardHan:
@dwalintukan if no scroll, we can use scroll='body', so when the height is not enough, the user will scroll the whole dialog instead of scrolling inside the dialog. The margin gets all wrong tho.
https://material-ui.com/demos/dialogs/
Check the last demo.

Do you want me to do it?

dwalintukan:
Ya I played around with it before. If you can get the margins right with scroll body I say go for it. 

This thread talks about it:
https://github.com/mui-org/material-ui/issues/10000

Native CSS trick:
https://stackoverflow.com/a/16671476/2917742

PR #1047, opened 2018-10-09
seperate loading dialog and create dialog

BernardHan:
Tried overflow: hidden, but the scrollbar is still there. I want to put the hiding on hold since other tasks are more important.

#1036

### DIFF
--- a/src/scenes/CreateEvent/index.js
+++ b/src/scenes/CreateEvent/index.js
@@ -28,51 +28,59 @@ const messages = defineMessages({
   },
 });
 
-const CreateEventDialog = ({ classes, store: { createEvent: { isOpen, loaded } } }) => (
+const CreateEventDialog = () => (
+  <Fragment>
+    <Loading />
+    <CreateEventDetail />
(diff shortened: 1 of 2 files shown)


dwalintukan:
Prefer not to do a different dialog just for loading:
https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Dialog/Dialog.js

Look at the source code styles. If you apply specific scrollbar styles you should be able to hide the scrollbar. As I mentioned, we should be be able to scroll the dialog, but just hide the bar UI element.

BernardHan:
@dwalintukan I fixed the create event dialog width, and made loading and dialog use the same `Dialog` element.

As for scroll bar, I tried to read the source code, and tried to find a way in Chrome element tool (change properties around each level), but no good result for now.

I feel like it's not a simple change on the CSS. So I'm creating an issue for this, and we can revisit later. For now, I don't think it's important.

dwalintukan:
@BernardHan That's ok with me. This is not that big of an issue.

Issue #967, opened 2018-09-21
Pull transaction list based on Votes, Set Results, Finalizes, Withdraws


dwalintukan:
Currently the `Transactions` table in the Event Detail is pulling off the `Transactions DB`. This is not good since it is not based off the blockchain data. Transactions is a model we created ourselves.

We should be pulling off the `Votes, Set Results, Finalizes, Withdraws` tables instead. These are all confirmed by the blockchain.

Will require some changes to the backend.

BernardHan:
This issue can get started after server issue 144 finished.
https://github.com/bodhiproject/bodhi-server/issues/144

PR #1082, opened 2018-10-26
967 txlist

BernardHan:
#967 

For now I have finished pulling history of
1. pending transactions
2. bets
3. result set
4. votes

I added a model file called `Vote.js` for processing some data necessary for `txRow.js` in history table.

Please take a look, if it's ok I will do the same to finalizes and withdraws.

For withdraw history, when it's a pending transaction, it shows type of `withdraw`, but in db it actually has either `escrow` or `winning` and amount.
Do you want the confirmed history to show the same as pending? Or as db shows? I root for the latter.


src/constants.js
```diff
@@ -94,4 +94,10 @@ module.exports = {
     CREATE_EVENT: 3500000,
     DORACLE_CREATE: 1500000,
   },
+
+  DBVoteType: {
```
dwalintukan: just `VoteType` is better since we are using the same type for both backend/UI. better to keep it consistent across all stacks.

src/network/graphql/queries.js
```diff
@@ -337,3 +338,37 @@ export async function searchTopics(app, phrase, filters, orderBy) {
   const result = await request.execute();
   return map(result, (topic) => new Topic(topic, app));
 }
+
+/**
+ * Search for Topics that contains phrase either in title or result setter address
```
dwalintukan: copy/paste?

src/network/graphql/queries.js
```diff
@@ -337,3 +338,37 @@ export async function searchTopics(app, phrase, filters, orderBy) {
...
+ * @param {Array} filters Array of objects for filtering. ie. [{ status: 'WAITRESULT' }, { status: 'OPENRESULTSET' }]
+ * @param {Object} orderBy Object with order by fields. ie. { field: 'blockNum', direction: 'DESC' }
+ * @return {Promise} Search result from graphql
```
dwalintukan: wrong desc for return

src/network/graphql/queries.js
```diff
@@ -337,3 +338,37 @@ export async function searchTopics(app, phrase, filters, orderBy) {
...
+ */
+export function queryResultSets(filters, orderBy) {
+  const request = new GraphQuery('resultSets', TYPE.resultSet);
+  if (!isEmpty(filters)) {
+    request.setFilters(filters);
+  }
+  if (!isEmpty(orderBy)) {
+    request.setOrderBy(orderBy);
+  }
+  return request.execute();
```
dwalintukan: create `ResultSet` model and return mapped array here

src/network/graphql/queries.js
```diff
@@ -337,3 +338,37 @@ export async function searchTopics(app, phrase, filters, orderBy) {
...
+}
+
+/**
+ * Search for Topics that contains phrase either in title or result setter address
```
dwalintukan: change desc

src/network/graphql/queries.js
```diff
@@ -337,3 +338,37 @@ export async function searchTopics(app, phrase, filters, orderBy) {
...
+ * @param {Array} filters Array of objects for filtering. ie. [{ status: 'WAITRESULT' }, { status: 'OPENRESULTSET' }]
+ * @param {Object} orderBy Object with order by fields. ie. { field: 'blockNum', direction: 'DESC' }
+ * @return {Promise} Search result from graphql
```
dwalintukan: change return desc

src/network/graphql/queries.js
```diff
@@ -337,3 +338,37 @@ export async function searchTopics(app, phrase, filters, orderBy) {
...
+ */
+export function queryWithdraws(filters, orderBy) {
+  const request = new GraphQuery('withdraws', TYPE.withdraw);
+  if (!isEmpty(filters)) {
+    request.setFilters(filters);
+  }
+  if (!isEmpty(orderBy)) {
+    request.setOrderBy(orderBy);
+  }
+  return request.execute();
```
dwalintukan: create `Withdraw` model and return mapped array here

src/network/graphql/schema.js
```diff
@@ -112,13 +114,23 @@ const TYPE_DEF = {
...
+    }
     voterAddress
     topicAddress
     oracleAddress
     optionIdx
     amount
     version
+    token
+    type
+    topic {
```
dwalintukan: as discussed, remove `topic` from `vote` obj

src/scenes/Event/store.js
```diff
@@ -19,7 +19,12 @@ const INIT = {
   loading: true,
   oracles: [],
   topics: [],
+  leaderboardVotes: [],
   votes: [],
+  bets: [],
+  resultsets: [],
+  finalizes: [],
+  withdraws: [],
```
dwalintukan: i'm thinking we can just use one single array (`transactionHistoryItems`) for all: `transactions, bets, votes, resultSets, finalizes, withdraws` unless you have a good reason to keep them separated? and we just sort the entire list based on shared fields e.g. `blockTime`.

dwalintukan: oh looks like `this.transactions` is what i described. nit: to rename to `transactionHistoryItems` to be clear

src/scenes/Event/store.js
```diff
@@ -295,10 +305,33 @@ export default class EventStore {
...
+    // // );
+
+    // // this.finalizes = await queryResultSets(
+    // //   [{ topicAddress: address, oracleAddress: null }],
+    // //   { field: 'block.blockTime', direction: SortBy.DESCENDING },
+    // // );
+
+    const votes = await queryAllVotes(
+      [{ topicAddress: address }],
+      { field: 'block.blockTime', direction: SortBy.DESCENDING },
```
dwalintukan: dont need to do any sorting (can remove the orderBy) until all the items are in one array since we want to sort it by a shared field among all the items like `blockTime`

src/stores/models/Vote.js
```diff
@@ -0,0 +1,42 @@
+import { TransactionStatus } from 'constants';
+
+import { satoshiToDecimal, decimalToSatoshi } from '../../helpers/utility';
+
+/**
+ * Model for Transactions.
```
dwalintukan: Votes

src/stores/models/Vote.js
```diff
@@ -0,0 +1,42 @@
...
+  token = '' // Token type used for Transaction. QTUM for BET. BOT for VOTE.
+  version = 0 // Current version of the contract. To manage deprecations later.
+  localizedInvalid = {}; // for invalid option
+  topic
+
+  constructor(vote) {
+    Object.assign(this, vote);
+    this.status = TransactionStatus.SUCCESS;
+    this.amount = satoshiToDecimal(this.amount).toString();
+    this.amountSatoshi = decimalToSatoshi(this.amount);
```
dwalintukan: can save computation by setting `amountSatoshi` first as `this.amount` then `this.amount = satoshiToDecimal(this.amountSatoshi).toString();`

src/stores/models/Vote.js
```diff
@@ -0,0 +1,42 @@
...
+  voterAddress = '' // Sender's address
+  topicAddress = '' // Topic contract address associated with Transaction
+  oracleAddress = '' // Oracle contract address associated with Transaction
+  optionIdx = 0 // Result index used for Transaction. eg. For a bet, this would be the result index the user bet on.
+  amount = '' // Token amount in decimals
+  amountSatoshi = '' // Token amount in satoshi
+  token = '' // Token type used for Transaction. QTUM for BET. BOT for VOTE.
+  version = 0 // Current version of the contract. To manage deprecations later.
+  localizedInvalid = {}; // for invalid option
+  topic
```
dwalintukan: lets remove `topic` and `localizedInvalid` from the Vote model. and you can pull that from the Topic obj in the Event store.

src/stores/models/Vote.js
```diff
@@ -0,0 +1,42 @@
...
+import { satoshiToDecimal, decimalToSatoshi } from '../../helpers/utility';
+
+/**
+ * Model for Transactions.
+ */
+export default class Vote {
+  type = '' // One of: DBVoteType
+  txid = '' // Transaction ID assigned by the blockchain
+  status = '' // only success
+  block
```
dwalintukan: don't need. blockNum and blockTime handles the obj.

src/stores/models/Vote.js
```diff
@@ -0,0 +1,42 @@
...
+  blockNum = 0 // Block number when Transaction was recorded on blockchain
+  blockTime = '' // Block timestamp for blockNum
+  voterAddress = '' // Sender's address
+  topicAddress = '' // Topic contract address associated with Transaction
+  oracleAddress = '' // Oracle contract address associated with Transaction
+  optionIdx = 0 // Result index used for Transaction. eg. For a bet, this would be the result index the user bet on.
```
dwalintukan: set this as `null` so we don't confuse an empty Vote obj vs an errored Vote.optionIdx.

src/stores/models/Vote.js
```diff
@@ -0,0 +1,42 @@
...
+
+import { satoshiToDecimal, decimalToSatoshi } from '../../helpers/utility';
+
+/**
+ * Model for Transactions.
+ */
+export default class Vote {
+  type = '' // One of: DBVoteType
+  txid = '' // Transaction ID assigned by the blockchain
+  status = '' // only success
```
dwalintukan: lets not add this field to the model itself, but add it to the UI logic. someone might confuse that the DB Vote model has a `status` field when it doesn't so lets just fill in the `Success` field in table if its not a `Transaction` type.

src/scenes/Event/store.js
BernardHan: Absolutely sure. The only reason I made these arrays is in case in the future we need to use any specific one of them. It's my personal coding habit, not sure if it's a good one tho...

src/network/graphql/queries.js
BernardHan: Oh I meant to change those in the last step before push my final version

src/network/graphql/queries.js
BernardHan: Those were prototype, haven't got started on resultset and withdraws

src/stores/models/Vote.js
BernardHan: for data from `transactions.db`, do we display it if it's status of fail?

src/scenes/Event/store.js
dwalintukan: Ya just use the single array for now. If we have a use for using individual ones later it will be a quick change to track each one.

src/stores/models/Vote.js
dwalintukan: Yes we do show failed txs as well.

src/scenes/Event/store.js
```diff
@@ -295,15 +306,28 @@ export default class EventStore {
 
   @action
   queryTransactions = async (address) => {
-    this.transactions = await queryAllTransactions(
-      [{ topicAddress: address }],
+    const pendings = await queryAllTransactions(
+      [{ topicAddress: address, status: TransactionStatus.PENDING }],
       { field: 'createdTime', direction: SortBy.DESCENDING },
     );
+    this.withdraws = await queryWithdraws([{ topicAddress: address }]);
```
dwalintukan: oh thought you were going to remove the individual arrays for each type? are the individual arrays used somewhere? if not, I wouldn't bother storing them.

src/scenes/Event/store.js
```diff
@@ -295,15 +306,28 @@ export default class EventStore {
...
+
+    this.finalizes = await queryFinalizes([{ topicAddress: address }]);
+    this.finalizes = filter(this.finalizes, { oracleAddress: null });
```
dwalintukan: do you need to pass in null as `oracleAddress` or can you just leave out the filter entirely?

src/stores/models/Finalize.js
```diff
@@ -0,0 +1,21 @@
...
+  fromAddress = '' // Sender's address
+  topicAddress = '' // Topic contract address associated with Finalize
+  optionIdx = null // Result index used for Finalize. eg. For a bet, this would be the result index the user bet on.
+  version = 0 // Current version of the contract. To manage deprecations later.
+
+  constructor(finalize) {
+    Object.assign(this, finalize);
+    this.blockTime = finalize.block.blockTime;
+    this.blockNum = finalize.block.blockNum;
+    this.type = 'FINALIZERESULT';
```
dwalintukan: RESULTSET?

src/scenes/Event/store.js
BernardHan: I will to remove them eventually.

src/scenes/Event/store.js
BernardHan: `oracleAddress == null` is the rule to say if it's `finalize`, right? The result from the query would give both `resultset` and `finalize` belong to that `topicAddress`, I only want to get the `finalize` tho.

dwalintukan: Yes you are right. Was wondering why the finalize doesn't have an oracleAddress so I checked the contract code and its cause it is executed in the TopicEvent. Bad design on my part when I created the contracts, should've had separate events for ResultSet and Finalize.

src/stores/models/Finalize.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Model for Finalize.
+ */
+export default class Finalize {
```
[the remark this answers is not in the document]
dwalintukan: I know why you used `Finalize` instead of `ResultSet`, but we should still use the name of the DB layer model `ResultSet`. If we ever used `ResultSet` later on, we would use the same model so the Finalize model does not fit the naming for it.

src/scenes/Event/store.js
```diff
@@ -295,15 +296,29 @@ export default class EventStore {
 
   @action
   queryTransactions = async (address) => {
-    this.transactions = await queryAllTransactions(
-      [{ topicAddress: address }],
+    const pendings = await queryAllTransactions(
+      [{ topicAddress: address, status: TransactionStatus.PENDING }, { topicAddress: address, status: TransactionStatus.FAIL }],
```
dwalintukan: nice

src/stores/models/ResultSet.js
```diff
@@ -0,0 +1,24 @@
...
+  txid = '' // ResultSet ID assigned by the blockchain
+  blockNum = 0 // Block number when ResultSet was recorded on blockchain
+  blockTime = '' // Block timestamp for blockNum
+  fromAddress = '' // Sender's address
+  topicAddress = '' // Topic contract address associated with ResultSet
+  oracleAddress = ''
+  optionIdx = null // Result index used for ResultSet. eg. For a bet, this would be the result index the user bet on.
+  version = 0 // Current version of the contract. To manage deprecations later.
+
+  constructor(resultset) {
```
dwalintukan: nit: resultSet

test/mockDB.js
```diff
@@ -47,6 +47,9 @@ export default {
   paginatedOracles: INIT_VALUES.paginatedOracles,
   paginatedAccumulatedVotes: INIT_VALUES.paginatedAccumulatedVotes,
   transactions: [],
+  withdraws: [],
+  resultsets: [],
```
dwalintukan: nit: resultSets

src/scenes/Event/store.js
```diff
@@ -295,15 +296,29 @@ export default class EventStore {
...
       { field: 'createdTime', direction: SortBy.DESCENDING },
     );
+    const withdraws = await queryWithdraws([{ topicAddress: address }]);
+
+    const resultsets = await queryResultSets([{ topicAddress: address }]);
```
dwalintukan: nit: resultSets

src/scenes/Event/store.js
```diff
@@ -295,15 +296,29 @@ export default class EventStore {
...
+    const finalizes = filter(resultsets, { oracleAddress: null });
+
+    const transactions = await queryAllVotes([{ topicAddress: address }]);
+    const votes = filter(transactions, { type: TransactionType.VOTE });
+    const resultsetsWithAmount = filter(transactions, { type: TransactionType.SET_RESULT });
```
dwalintukan: nit: resultSetsWithAmount

PR #1092, opened 2018-11-09
theme font

GoodThomas:


src/config/theme.js
```diff
@@ -4,14 +4,15 @@ import { createMuiTheme } from '@material-ui/core';
...
-// Font size
-const fontSizeTitleLg = 36;
-const fontSizeTitleMd = 32;
-const fontSizeTitleSm = 24;
-const fontSizeTextLg = 20;
-const fontSizeTextMd = 18;
-const fontSizeTextSm = 16;
-const fontSizeMeta = 14;
+// Type scale
+const xxxSmallText = 10;
```
dwalintukan: let's use `rem` units for everything

src/config/theme.js
```diff
@@ -134,41 +135,66 @@ export const theme = {
   typography: {
     useNextVariants: true,
     fontFamily: fontLato,
-    fontSize: fontSizeTextSm,
+    fontSize: px(xSmallText),
```
dwalintukan: should be `smallText`. fontSizeTextSm = 16

src/config/theme.js
```diff
@@ -350,14 +401,14 @@ const bodhiTheme = createMuiTheme(theme);
...
-    lineHeight: lineHeightLg,
+  subtitle1: {
+    fontSize: px(mediumText),
+    fontWeight: fontWeightBold,
+    lineHeight: lineHeightSm,
     marginLeft: '0',
     color: textColorDarkGrey,
     [bodhiTheme.breakpoints.down('xs')]: {
-      fontSize: fontSizeTitleSm,
+      fontSize: px(largeText),
```
dwalintukan: `smallText`?

src/config/theme.js
dwalintukan: 1. set the global fontSize to 16px (1rem)
2. change to use rem everywhere

```
const smallText = 1rem;
const mediumText = 1.25 rem;
etc
```

PR #1093, opened 2018-11-09
Theme button

GoodThomas:


src/config/theme.js
```diff
@@ -395,39 +397,88 @@ bodhiTheme.overrides = {
   ...bodhiTheme.overrides,
   MuiButton: {
     root: {
+      fontSize: xSmallText,
```
dwalintukan: should be 1rem, smallText

src/config/theme.js
```diff
@@ -395,39 +397,88 @@ bodhiTheme.overrides = {
...
+      lineHeight: lineHeight24,
+      padding: `4px ${px(space2X)}`,
```
dwalintukan: where did you get these padding values from?

src/config/theme.js
```diff
@@ -395,39 +397,88 @@ bodhiTheme.overrides = {
...
+      '&:active': {
+        backgroundColor: primaryBrandColor60,
+        boxShadow: 'none',
       },
       '&:disabled': {
-        backgroundColor: primaryBrandColor,
-        color: white,
-        opacity: 0.6,
+        backgroundColor: primaryBrandColor30,
+        color: `${white}!important`,
```
dwalintukan: add space in between

src/config/theme.js
```diff
@@ -395,39 +397,88 @@ bodhiTheme.overrides = {
...
+        boxShadow: 'none',
+      },
+    },
+    outlined: {
+      color: primaryBrandColor,
+      border: `1px solid ${primaryBrandColor}`,
```
dwalintukan: it looks like the design has it as 2px?

src/config/theme.js
```diff
@@ -395,39 +397,88 @@ bodhiTheme.overrides = {
...
+        boxShadow: `1px 1px 4px ${primaryBrandColor60}`,
+      },
+      '&:disabled': {
+        color: primaryBrandColor30,
+        boxShadow: 'none',
       },
     },
     sizeLarge: {
-      fontSize: mediumText,
+      fontSize: smallText,
```
dwalintukan: according to the design, large buttons are 24px = `largeText`

src/config/theme.js
```diff
@@ -395,39 +397,88 @@ bodhiTheme.overrides = {
...
+      lineHeight: lineHeight24,
+      padding: `12px ${px(space2X)}`,
```
dwalintukan: where is this padding from?

src/config/theme.js
```diff
@@ -395,39 +397,88 @@ bodhiTheme.overrides = {
...
       fontWeight: fontWeightBold,
-      minHeight: 56,
+      minWidth: 'auto',
+      minHeight: 'auto',
     },
     sizeSmall: {
       fontSize: xxSmallText,
+      lineHeight: lineHeight16,
+      padding: `4px ${px(space2X)}`,
```
dwalintukan: where is the padding from?

src/config/theme.js
```diff
@@ -395,39 +397,88 @@ bodhiTheme.overrides = {
...
     },
     sizeLarge: {
-      fontSize: mediumText,
+      fontSize: smallText,
+      lineHeight: lineHeight24,
+      padding: `12px ${px(space2X)}`,
       fontWeight: fontWeightBold,
-      minHeight: 56,
+      minWidth: 'auto',
+      minHeight: 'auto',
```
dwalintukan: do we need to set min width/height?

src/config/theme.js
```diff
@@ -395,39 +397,88 @@ bodhiTheme.overrides = {
...
     },
     sizeSmall: {
       fontSize: xxSmallText,
```
dwalintukan: design says 16px = 1rem

GoodThomas:
It seems to be no `sizeMedium` property, so I overrode styles in `root` property.

src/config/theme.js
GoodThomas: I set these is to override the default values.

src/config/theme.js
GoodThomas: It seems to be no `sizeMedium` property, so I overrode styles in `root` property.
Medium size button seems to be 14px.

src/config/theme.js
GoodThomas: 16px
image
4px = (height 32px - line-height 24px) / 2

src/config/theme.js
GoodThomas: space3X = 24px
image
12px = (height 48px - line-height 24px) / 2

src/config/theme.js
GoodThomas: space2X = 16px
image
4px = (height 24px - line-height 16px) / 2

src/config/theme.js
GoodThomas: image

src/config/theme.js
dwalintukan: you are right, medium is the default size so root seems like the right spot. and for the font size i was mistaking the line height as the font size... sorry

src/config/theme.js
dwalintukan: you are right. I mistook line height for font size.

PR #1094, opened 2018-11-09
Theme form

GoodThomas:


src/config/theme.js
```diff
@@ -281,6 +281,35 @@ export const theme = {
...
+          borderBottom: `2px solid ${primaryBrandColor}`,
+        },
+      },
+      error: {
+        '&:after': {
+          borderBottomColor: `${supportColorRed100} !important`,
+        },
+      },
+    },
+    MuiFormHelperText: {
```
dwalintukan: does this apply to all `MuiInput`s? or do we have to wrap all input fields with `FormHelper`?

GoodThomas: The TextField contains a FormHelperText. If use Input, it needs FormHelperText. In this project, using TextField with FormHelperText, it doesn't matter.

PR #1095, opened 2018-11-13
theme button hover

GoodThomas:


src/config/theme.js
```diff
@@ -417,50 +420,112 @@ bodhiTheme.overrides = {
       color: white,
       boxShadow: 'none',
       '&:hover': {
-        backgroundColor: primaryBrandColor80,
+        backgroundColor: primaryBrandColorTransparent80,
```
dwalintukan: can we just use the `opacity: 0.8` field here?

src/config/theme.js
```diff
@@ -417,50 +420,112 @@ bodhiTheme.overrides = {
...
         boxShadow: 'none',
       },
       '&:active': {
-        backgroundColor: primaryBrandColor60,
+        backgroundColor: primaryBrandColorTransparent60,
```
dwalintukan: can we use `opacity: 0.6`?

src/config/theme.js
GoodThomas: In some places, the transparency of backgroundColor changes, but the transparency of text and border doesn't change.
image

PR #1141, opened 2019-05-14
Arbitration

BernardHan:
#1125

Previous `OracleFilter` has a field called `excludeResultSetterAddress`, which is used to hide `Result_Waiting` events from those events owner in `BotCourt` page.

Now since we don't have such field in current schema, I will just make all `Arbitration` and `ORACLE_RESULT_SETTING` status events show up.

### DIFF
--- a/src/scenes/Arbitration/store.js
+++ b/src/scenes/Arbitration/store.js
@@ -1,7 +1,7 @@
 import { observable, action, runInAction, reaction, toJS } from 'mobx';
 import _ from 'lodash';
-import { Token, OracleStatus, Routes } from '../../constants';
-import { queryAllOracles } from '../../network/graphql/queries';
+import { EVENT_STATUS, Routes } from '../../constants';
+import { events } from '../../network/graphql/queries';
 
 const INIT_VALUES = {
   loaded: false, // loading state?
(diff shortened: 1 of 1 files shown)


dwalintukan:
@BernardHan Sorry I will add this back in.

src/scenes/Arbitration/store.js
```diff
@@ -69,21 +69,23 @@ export default class ArbitrationStore {
...
+        },
+      ];
+
+      const { account } = this.app.naka;
+      if (!account) await window.ethereum.enable();
+
+      const res = await events({ filters, orderBy, limit, skip, pendingTxsAddress: account });
+      if (res.pageInfo) this.hasMore = res.pageInfo.hasNextPage;
+      else this.hasMore = false;
+      return _.orderBy(res.items, ['endTime'], this.app.sortBy.toLowerCase());
```
dwalintukan: please import only what you need from lodash.

src/scenes/Arbitration/store.js
```diff
@@ -69,21 +69,23 @@ export default class ArbitrationStore {
...
-    //       excludeResultSetterAddress,
-    //       language: this.app.ui.locale,
-    //     },
-    //   ];
-    //   const { oracles, pageInfo: { hasNextPage } } = await queryAllOracles(this.app, filters, orderBy, limit, skip);
-    //   this.hasMore = hasNextPage;
-    //   return _.orderBy(oracles, ['endTime'], this.app.sortBy.toLowerCase());
-    // }
+    if (this.hasMore) {
+      const orderBy = { field: 'endTime', direction: this.app.sortBy };
```
dwalintukan: `endTime` is not longer a field. use `arbitrationEndTime`.

src/scenes/Arbitration/store.js
dwalintukan: change `endTime` -> `arbitrationEndTime`

dwalintukan: nit: change using `this.app.sortBy.toLowerCase()` to `constants.SortBy.DESCENDING.toLowerCase()`. not sure why we have a global sort by set in the app store?

PR #1155, opened 2019-05-16
Support create event gas using nbot

henryzbw:


src/stores/WalletStore.js
```diff
@@ -46,6 +49,8 @@ export default class WalletStore {
   onNakaAccountChange = async (account) => {
     const { loggedIn, network, address, balance } = account;
 
+    this.nbotMethods = window.naka.eth.contract(getContracts().NakaBodhiToken.abi).at(getContracts().NakaBodhiToken.address);
```
dwalintukan: nit: `nbotContract` probably makes more sense as a name here since you are creating an instance of `contract` at the address.

in web3 1.0, after you create the `Contract` instance, you have to call `contract.methods.myMethod()` which is why I created an `nbotMethods` to reference `contract.methods`. but in web3 < 1.0, the methods are at the same level as the contract functions.

dwalintukan: create default class variable: `nbotMethods = undefined` so we are aware of what vars can be accessed by the class.

src/stores/TransactionStore.js
```diff
@@ -117,9 +117,15 @@ export default class TransactionStore {
         createEventFuncTypes,
         eventParams,
       ).substr(2);
+
       const data = `0x${CREATE_EVENT_FUNC_SIG}${paramsHex}`;
       // Send tx
-      const txid = await promisify(nbotMethods.transfer['address,uint256,bytes'].sendTransaction, [eventFactoryAddr, escrowAmt, data, { gas }]);
+      const txid = await promisify(nbotMethods.transfer['address,uint256,bytes'].sendTransaction, [eventFactoryAddr, escrowAmt, data, {
```
dwalintukan: add validation before calling this function. if we dont have a valid exchanger or exchangeRate then we should show an error message.

PR #1166, opened 2019-05-20
1134 search

dwalintukan:
#1134 
- some fav drawer fixes also


src/scenes/Search/store.js
```diff
@@ -1,93 +1,85 @@
...
-          this.bets = [];
-          this.votes = [];
-          this.sets = [];
-        }
-      },
-    );
+
     reaction(
-      () => this.app.global.syncBlockNum + this.app.global.online,
+      () => this.app.global.online,
```
BernardHan: previously fetch on new block in case the search result changes event status.

now that new block comes out too quickly, guess we can remove it for now.

PR #1178, opened 2019-05-21
1177 replace peers

dwalintukan:
#1177 
- replace peer count
- move wallet status to bottom bar

image

image

### DIFF
--- a/src/components/BottomBar/index.js
+++ b/src/components/BottomBar/index.js
@@ -20,11 +20,18 @@ export default class BottomBar extends Component {
   }
 
   render() {
-    const { classes, width } = this.props;
-    const { syncBlockTime, syncBlockNum, online } = this.props.store.global;
+    const {
+      classes,
+      width,
+      store: {
+        global: { syncBlockTime, syncBlockNum, online },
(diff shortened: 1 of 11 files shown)


src/components/BottomBar/index.js
```diff
@@ -21,59 +20,129 @@ export default class BottomBar extends Component {
...
+            ? <CheckCircle className={cx(classes.statusIcon, 'online')} />
+            : <RemoveCircle className={cx(classes.statusIcon, 'offline')} />
+          }
+          <Typography variant="body2" className={classes.bottomBarTxt}>
+            <FormattedMessage id="bottomBar.wallet" defaultMessage="Wallet" />
+          </Typography>
+        </div>
+      </Fragment>
+    )}
+    {width === 'xs' && (
```
henryzbw: Does material ui have grid system can directly apply responsive page instead of having the code duplicate?

henryzbw:
image
And the mobile version is a little off.

src/components/BottomBar/index.js
dwalintukan: nope

PR #1180, opened 2019-05-21
Fix favorite card styling

dwalintukan:
image


src/components/RaisedAmount/index.js
```diff
@@ -0,0 +1,37 @@
...
+import styles from './styles';
+
+@injectIntl
+@withStyles(styles, { withTheme: true })
+@inject('store')
+@observer
+export default class RaisedAmount extends Component {
+  static propTypes = {
+    classes: PropTypes.object.isRequired,
+    amount: PropTypes.string.isRequired,
```
henryzbw: the type should be number

PR #1201, opened 2019-05-23
1199 version switch

dwalintukan:
#1199 

image


src/stores/TransactionStore.js
```diff
@@ -61,27 +77,19 @@ export default class TransactionStore {
...
       // Send tx
+      const pbtParams = this.getPayByTokenParams();
       const txid = await promisify(
         nbotMethods.transfer['address,uint256,bytes'].sendTransaction,
         [eventFactoryAddr, escrowAmt, data, {
-          token: getContracts().NakaBodhiToken.address,
-          exchanger: nbotOwner,
-          exchangeRate,
           gas,
+          ...pbtParams,
```
BernardHan: `pbtParams` is an object, don't think we can use spread syntax on object?

dwalintukan: yup it works on objects as well

PR #1214, opened 2019-05-24
Fix gas fee formula

BernardHan:
`satoshiToDecimal(cumulativeGasUsed)` returns 0, hence the gasfee would be 0. Not sure if this is correct.

### DIFF
--- a/src/scenes/Activities/ActivityHistory/EventRows/index.js
+++ b/src/scenes/Activities/ActivityHistory/EventRows/index.js
@@ -59,10 +59,10 @@ class EventRow extends Component {
 
     render() {
       const { transaction, intl, classes, store: { wallet: { exchangeRate } } } = this.props;
-      const { txType, txid, txReceipt: { cumulativeGasUsed, gasPrice }, txStatus, block, name, address, amount } = transaction;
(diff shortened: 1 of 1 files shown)


src/scenes/Activities/ActivityHistory/EventRows/index.js
```diff
@@ -58,10 +58,11 @@ class EventRow extends Component {
     }
 
     render() {
-      const { transaction, intl, classes } = this.props;
-      const { txType, txid, txReceipt: { cumulativeGasUsed }, txStatus, block, name, address, amount } = transaction;
+      const { transaction, intl, classes, store: { wallet: { exchangeRate } } } = this.props;
+      const { txType, txid, txReceipt: { cumulativeGasUsed, gasPrice }, txStatus, block, name, address, amount } = transaction;
       const blockTime = block ? block.blockTime : messages.strPendingMsg;
       const { expanded } = this.state;
+      const gasFee = (satoshiToDecimal(gasPrice) / weiToDecimal(exchangeRate)) * satoshiToDecimal(cumulativeGasUsed);
```
dwalintukan: if the tx is unconfirmed, cumulativeGasUsed will be undefined but gasUsed will be defined. 

If tx is confirmed, you can use cumulative.

PR #1218, opened 2019-05-24
Fix dup history

BernardHan:


dwalintukan:
After the backend `transactions` query is changed the UI shouldnt have to do any changes for it to work.

BernardHan:
@dwalintukan This is to compat with the `fromVote` in server side. Or do you just want the server to filter when executing `transactions` by comparing one by one to find the duplicates?

dwalintukan:
@BernardHan Correct. We shouldn't actually remove items from the DB as that would have differing items from DB and chain. The only query this affects is the `transactions` query so we need to run the query then remove the entries we dont need. After this is done on the server PR, then there won't be any need for this issue.

PR #1294, opened 2019-05-31
Change arbitration length

BernardHan:


src/scenes/CreateEvent/styles.js
```diff
@@ -55,6 +55,17 @@ const styles = (theme) => ({
   cancelButton: {
     marginRight: theme.padding.spaceX.px,
   },
+  card: {
```
dwalintukan: should this be applied everywhere? if so, might want to add this to the theme overrides.

BernardHan: Looks like henry and I split into different places, and different places have different effect. I think it's better to not add to theme for now.

PR #1298, opened 2019-05-31
Fix decimal amount

henryzbw:


src/helpers/utility.js
```diff
@@ -55,9 +55,13 @@ export function decimalToSatoshi(number) {
   if (!number) {
     return number;
   }
+  const toStringNumber = String(number);
+  const splitArr = toStringNumber.split('.');
+  const integerPart = splitArr[0];
+  const decimalPart = Number(`.${splitArr[1]}`) * SATOSHI_CONVERSION;
```
dwalintukan: what happens if there is no decimal?

src/helpers/utility.js
```diff
@@ -78,7 +83,10 @@ export function satoshiToDecimal(number) {
   }
 
   const conversionBN = toBN(SATOSHI_CONVERSION);
-  return bn.div(conversionBN).toNumber();
+  const integerPart = bn.div(conversionBN).toNumber();
+  const decimalPartInBN = bn.sub(conversionBN.mul(toBN(integerPart)));
```
dwalintukan: decimalPartBN to keep consistency

PR #1300, opened 2019-05-31
Replace qtum

henryzbw:


src/languageProvider/locales/en_US.json
```diff
@@ -413,9 +418,9 @@
   "withdrawDetail.returnRate": "Return rate:",
   "withdrawDetail.returnRateXPercent": "Return rate: {rate}%",
   "withdrawDetail.reward": "REWARD",
-  "withdrawDetail.totalBetTotalVote": "Total bet amount {naka} NAKA. Total voted amount {nbot} NBOT.",
+  "withdrawDetail.totalBetTotalVote": "Total bet amount {totalBets} NBOT. You bet amount {youBet} NBOT.",
```
dwalintukan: is this one used or `withdrawDetail.youBetYouVote`? seems like its used for the same purpose.

henryzbw: seems both used, will check all the strings when test the whole app.

PR #1387, opened 2019-06-06
Transaction loadmore

BernardHan:
#1361 

This pr is changes to adapt to the server changes related to transaction query.

Here I'm moving all history related stuff into `store/HistoryStore`, which can be used by `my activity history` and `event all transaction, my transaction and result history`


src/network/graphql/queries.js
```diff
@@ -168,11 +168,13 @@ const QUERIES = {
       $filter: TransactionFilter
       $limit: Int
       $skip: Int
+      $skips: TransactionSkip
```
dwalintukan: NextTransactionSkip

src/network/graphql/queries.js
```diff
@@ -168,11 +168,13 @@ const QUERIES = {
...
     ) {
       transactions(
         filter: $filter
         limit: $limit
         skip: $skip
+        skips: $skips
```
dwalintukan: nextTransactionSkip: $nextTransactionSkip

src/network/graphql/schema.js
```diff
@@ -1,7 +1,15 @@
+export const NEXT_SKIPS = `
```
dwalintukan: NEXT_TRANSACTION_SKIP

src/network/graphql/schema.js
```diff
@@ -1,7 +1,15 @@
...
+  nextBetSkip
+  nextResultSetSkip
+  nextWithdrawSkip
+`;
+
 export const PAGE_INFO = `
   hasNextPage
   pageNumber
   count
+  nextSkips { ${NEXT_SKIPS} }
```
dwalintukan: nextTransactionSkip

src/scenes/Event/store.js
```diff
@@ -182,8 +182,8 @@ export default class EventStore {
     if (!this.event) return;
     this.address = this.event.address;
     this.escrowAmount = this.event.escrowAmount;
-    await this.queryResultSets();
-    await this.queryTransactions();
+    // await this.queryResultSets();
+    // await this.queryTransactions();
```
dwalintukan: remove if unused

src/stores/HistoryStore.js
```diff
@@ -0,0 +1,239 @@
...
+  myWithdrawSkip = INIT_VALUES.withdrawSkip;
+
+  constructor(app) {
+    this.app = app;
+
+    // Wallet addresses changed
+    reaction(
+      () => this.app.naka.account,
+      () => {
+        if (this.app.ui.location === Routes.ACTIVITY_HISTORY) {
```
dwalintukan: do you need to react on the Event detail page too?

src/stores/HistoryStore.js
```diff
@@ -0,0 +1,239 @@
...
+    // New block
+    reaction(
+      () => this.app.global.syncBlockNum,
+      async () => {
+        if (this.transactions.length > 0 && this.app.ui.location === Routes.ACTIVITY_HISTORY) {
+          const { naka: { account } } = this.app;
+          const filters = { transactorAddress: account };
+          const newTxs = await this.fetchHistory(filters, ACTIVITY_HISTORY_LIMIT, 0, 0, 0, 0, 0, true);
+          const old = this.transactions[0];
+          if (newTxs.length > 0 && (newTxs[0].txid !== old.txid
```
dwalintukan: why would this happen? if we are skipping the correct numbers it should always return the next item? sounds like a backend bug.

src/stores/HistoryStore.js
```diff
@@ -0,0 +1,239 @@
...
+      this.limit = EVENT_HISTORY_LIMIT;
+      await this.loadFirstTransactions({ eventAddress: address }); // for all txs
+
+      const myFilter = { eventAddress: address, transactorAddress: account };
+      this.myTransactions = await this.fetchMyHistory(myFilter); // for my txs
+    }
+  }
+
+  /**
+   * It is necessary to fetch a big batch of txs for the purposes of pagination. We need to know the total count
```
dwalintukan: update comments explaining passing in the tx skips

src/stores/HistoryStore.js
BernardHan: Good catch

src/stores/HistoryStore.js
BernardHan: Oh this piece of codes is not related to the loadmore transaction.

When user creates an event, it goes to `my activity history` page to let the user see he just did this transaction.
This is a temp solution to update the first tx (since the user can create event one at time) from pending to success without refreshing the whole page.

BernardHan: Do you have better idea? Since we also need similar solutions for event detail page, but that can be multiple txs.

dwalintukan: Ah I see. A more full-proof solution would prob be to do a find() on the transactions by txid and replace that index with the confirmed one. Can you always guarantee that it will be at index 0?

dwalintukan: Ya having multiple pending txs that need to be replaced should be done with a find() then cause I dont think the query will guarantee the index.

dwalintukan: I believe the pending is always at the top? So if thats the case, when you load more, you can check the pending items one by one, if you find that pending items in the confirmed items list, then you can remove it from the pending list.

PR #1404, opened 2019-06-07
Show previous consensus threshold

henryzbw:
image


src/scenes/Event/index.js
```diff
@@ -144,8 +148,11 @@ export default class EventPage extends Component {
 
   renderConsensusThresholdMessage = () => {
     const { intl, store: { eventPage } } = this.props;
-    const threshold = eventPage.event.consensusThreshold;
-    const heading = `${intl.formatMessage(messages.consensusThreshold)} ${threshold} NBOT`;
+    const { consensusThreshold, previousConsensusThreshold } = eventPage.event;
+    let heading = `${intl.formatMessage(messages.currentConsensusThreshold)} ${consensusThreshold} NBOT`;
+    if (eventPage.isArbitration) {
+      heading = `${heading}. ${intl.formatMessage(messages.previousConsensusThreshold)} ${previousConsensusThreshold} NBOT`;
```
dwalintukan: nit: add a line-break `\n` after the period to show on separate lines

dwalintukan: nit: also feel like the proper way would be to create a new string with both of them combined.

PR #1428, opened 2019-06-11
Show progress bar, fix #1417

### LINKED ISSUE #1417: all events page options not show progress bar percentage

henryzbw:
image


henryzbw:
Make user percentage and total percentage in eventcard
show previous round result index voting stuff
image

henryzbw:
fix #1423

src/models/Option.js
```diff
@@ -28,17 +29,24 @@ export default class Option {
     if (event.status !== EVENT_STATUS.ARBITRATION) {
       const totalBalance = _.sum(event.roundBets);
       this.percent = totalBalance === 0 ? totalBalance : _.round((this.amount / totalBalance) * 100);
+      this.userPercent = this.amount === 0 ? this.amount : _.round((event.userRoundBets[i] / this.amount) * this.percent);
     } else {
       this.isPrevResult = event.currentResultIndex === i;
+      this.value = this.isPrevResult ? event.previousRoundBets[i] : this.amount;
```
dwalintukan: why is there an `amount` and `value`? the naming seems so similar it's confusing to someone as to what each purpose is for.

src/models/Option.js
```diff
@@ -28,17 +29,24 @@ export default class Option {
...
       this.maxAmount = event.status === EVENT_STATUS.ARBITRATION
         ? event.consensusThreshold - decimalToSatoshi(this.amount) : undefined;
 
-      const threshold = this.isPrevResult ? 0 : event.consensusThreshold;
-      this.percent = threshold === 0 ? threshold : _.round((this.amount / threshold) * 100);
+      const threshold = this.isPrevResult ? event.previousConsensusThreshold : event.consensusThreshold;
+      this.percent = this.isPrevResult ? 100 : _.round((this.amount / threshold) * 100);
+      if (this.isPrevResult) {
```
dwalintukan: nit: inconsistent here with the way you set the values. `threshold` and `percent` using ternaries but checking the same `isPrevResult` while this one is an if statement. suggest to stick to one or the other

src/scenes/Event/Progress/styles.js
```diff
@@ -2,20 +2,21 @@
 
 const styles = (theme) => ({
   invalid: {
-    '& > div': {
+    '& > div:nth-child(2)': {
       'background-color': theme.palette.error.dark,
     },
+    '& > div:nth-child(3)': {
+      'background-color': '#e48181',
```
dwalintukan: set in theme and use. duplicated color code as the other Progress component.

src/scenes/Event/index.js
```diff
@@ -76,11 +77,18 @@ export default class EventPage extends Component {
...
+      store: {
+        eventPage: { event, event: { status, centralizedOracle } },
+        wallet: { currentAddress },
+      },
+    } = this.props;
     switch (status) {
       case EVENT_STATUS.CREATED: return 'creating';
-      case EVENT_STATUS.BETTING: return moment.unix().isBefore(moment.unix(event.betStartTime)) ? 'predictionComingSoon' : 'predictionInProgress';
-      case EVENT_STATUS.ORACLE_RESULT_SETTING: return moment.unix().isBefore(moment.unix(event.resultSetEndTime)) ? 'resultSettingComingSoon' : 'resultSettingInProgress';
+      case EVENT_STATUS.BETTING: return moment().isBefore(moment.unix(event.betStartTime)) ? 'predictionComingSoon' : 'predictionInProgress';
```
dwalintukan: seems to be exact same logic as the EventCard. good use case for helper func? or maybe ref same logic in the store?

src/components/Progress/styles.js
```diff
@@ -2,12 +2,22 @@
...
+    '& > div:nth-child(2)': {
       'background-color': theme.palette.error.dark,
     },
+    '& > div:nth-child(3)': {
+      'background-color': '#e48181',
+    },
+  },
+  root: {
+    height: '100% !important',
+    boxShadow: '0 2px 10px rgba(0,0,0,.2) !important',
```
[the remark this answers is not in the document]
henryzbw: no, there is a loader has different boxShadow

src/languageProvider/locales/zh-Hans.json
```diff
@@ -449,14 +434,12 @@
   "withdrawDetail.returnRate": "回报率：",
   "withdrawDetail.returnRateXPercent": "回报率：{rate}%",
   "withdrawDetail.reward": "回报",
-  "withdrawDetail.totalBetTotalVote": "Total bet amount {totalBets} NBOT. You bet amount {youBet} NBOT.",
   "withdrawDetail.withdrawTo": "提现到",
-  "withdrawDetail.youBetYouVote": "You bet {nbot} NBOT.",
+  "withdrawDetail.youBetYouVote": "您投注了{nbot}NBOT.",
```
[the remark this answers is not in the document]
henryzbw: no spaces needed for chinese

src/models/Option.js
henryzbw: use if statement here since linter requires not to nested ternaries

henryzbw: check the new one, it becomes more complex to handle the result set. And do i need to make all of the ternary to if statement?

src/models/Option.js
henryzbw: seems like the amount is only for temp use.

dwalintukan: hm can we get rid of it then?

src/models/Option.js
dwalintukan: I was just pointing out the code looks inconsistent by doing checks two diff ways:
```
myVar = isPrevResult ? a : b
myVar2 = isPrevResult ? a : b
if (isPrevResult) {
  a
} else {
  b
}
```

suggesting to do something like:
```
myVar
myVar2
if (isPrevResult) {
  myVar = a
  myVar2 = a
} else { 
  myVar = b
  myVar2 = b
}
```

not a huge deal though which is why i put nit. if you feel its ok, then just leave it.

Issue #1280, opened 2019-05-30
Change Date & Time pickers to Material UI


dwalintukan:
https://material-ui.com/components/pickers/#complementary-projects

I am currently updating Material UI to the latest 4.0.1 in my PR. Can handle this after my PR is merged.

PR #1436, opened 2019-06-12
1280 mat ui date time

dwalintukan:
#1280 
image


src/config/theme.js
```diff
@@ -306,13 +312,11 @@ export const theme = {
...
     },
     MuiFormHelperText: {
-      root: {
-        '&$error': {
-          fontSize: xxSmallText,
-          lineHeight: lineHeight16,
-          color: `${supportColorRed100} !important`,
-          marginTop: 0,
-        },
+      error: {
```
henryzbw:
```
You can not override it like this: 
{
  "error": {
    "marginTop": 4,
    "fontSize": "0.75rem",
    "lineHeight": "16px",
    "color": "#BE2828"
  }
}

Instead, you need to use the $ruleName syntax:
{
  "root": {
    "&$error": {
      "marginTop": 4,
      "fontSize": "0.75rem",
      "lineHeight": "16px",
      "color": "#BE2828"
    }
  }
}
```

dwalintukan: good to know, fixed

PR #1439, opened 2019-06-13
Handle display decimals

henryzbw:
image


src/helpers/utility.js
```diff
@@ -213,30 +215,10 @@ export function shortenAddress(text, maxLength) {
...
-      x = `0.${(new Array(e)).join('0')}${x.toString().substring(2)}`;
-    }
-  } else {
-    let e = parseInt(x.toString().split('+')[1], 10);
-    if (e > 20) {
-      e -= 20;
-      x /= 10 ** e;
-      x += (new Array(e + 1)).join('0');
-    }
+export function toFixed(num, isFullNumber) {
```
dwalintukan: can you do something like `toFixed(num, precision = 2)`? so that we dont need to pass an extra param when it seems like all of them are `true`.

also, it likes like `numeral(num.toFixed(precision)).format('0.00a');` can work for both cases?

PR #1463, opened 2019-06-14
Fix helper info not pop up in create event in mobile

BernardHan:
#1461


src/components/ImportantNote/index.js
```diff
@@ -1,22 +1,46 @@
...
+@injectIntl
+@withStyles(styles, { withTheme: true })
+@inject('store')
+@observer
+export default class ImportantNote extends Component {
+  state = {
+    open: false,
+  };
+
+  componentDidMount() {
```
henryzbw: Delete

PR #1466, opened 2019-06-14
Accomodate pre betting

henryzbw:


henryzbw:
wait for pre result setting

PR #1488, opened 2019-06-17
Add pre result setting

henryzbw:


src/helpers/utility.js
```diff
@@ -226,9 +226,12 @@ export function getEventDesc(event, currentAddress) {
     case EVENT_STATUS.CREATED: return 'creating';
     case EVENT_STATUS.PRE_BETTING: return 'predictionComingSoon';
     case EVENT_STATUS.BETTING: return 'predictionInProgress';
+    case EVENT_STATUS.PRE_RESULT_SETTING:
```
dwalintukan: fyi you can combine multiple case statements together
```
case EVENT_STATUS.PRE_RESULT_SETTING:
case EVENT_STATUS.ORACLE_RESULT_SETTING: {
      if (centralizedOracle !== currentAddress) return 'arbitrationComingSoon';
      return 'resultSettingComingSoon';
}
```

src/models/MultipleResultsEvent.js
```diff
@@ -73,7 +73,7 @@ export default class MultipleResultsEvent {
 
   isPending = () => Boolean(!!this.pendingTxs && this.pendingTxs.total && this.pendingTxs.total > 0);
 
-  isUpcoming = (address) => this.status === EVENT_STATUS.ORACLE_RESULT_SETTING
+  isUpcoming = (address) => [EVENT_STATUS.PRE_RESULT_SETTING, EVENT_STATUS.ORACLE_RESULT_SETTING].includes(this.status)
```
dwalintukan: can we rename `isUpcoming` to something more descriptive? is what upcoming?

src/scenes/Event/store.js
```diff
@@ -369,16 +369,15 @@ export default class EventStore {
...
+    if (status === PRE_RESULT_SETTING) {
       this.buttonDisabled = true;
       this.warningType = EventWarningType.INFO;
       this.eventWarningMessageId = 'oracle.setStartTimeDisabledText';
       return;
     }
 
     // User is not the result setter
-    if (status === ORACLE_RESULT_SETTING
+    if ([PRE_RESULT_SETTING, ORACLE_RESULT_SETTING].includes(status)
```
dwalintukan: Dont need to include `PRE_RESULT_SETTING` here. Should never hit here since the check on line 372.

src/helpers/utility.js
henryzbw: yeah, but  we cannot combine here, since if centralizedOracle === currentAddress and its oracle_result_setting, it should be result setting in progress

src/models/MultipleResultsEvent.js
henryzbw: actually, isupcoming is not used now, but we used to have it label oracle result setting event in arbitration tab

src/helpers/utility.js
dwalintukan: oops sorry didnt notice the strings were different.

src/models/MultipleResultsEvent.js
dwalintukan: can we remove it everywhere?

henryzbw: removed

PR #1491, opened 2019-06-17
Fix #1484 show full title on signle column

### LINKED ISSUE #1484: When the list page layout is single item, we should show full event title

henryzbw:
image
image

### DIFF
--- a/src/components/EventCard/index.js
+++ b/src/components/EventCard/index.js
@@ -92,34 +92,36 @@ export default class EventCard extends Component {
           <Card className={classes.eventCard} onClick={onClick}>
             <div className={cx(classes.eventCardBg, `bg${index % 8}`)}></div>
             <div className={cx(classes.eventCardSection, 'top')}>
-              {isPending() && status !== WITHDRAWING && <EventWarning id="str.pendingConfirmation" message="Pending Confirmation" />}
-              {isPending() && status === WITHDRAWING && <EventWarning id="str.withdrawing" message="Withdrawning" type={EventWarningType.INFO} />}
-              {this.isWithdrawn && <EventWarning id="str.withdrawn" message="Withdrawn" type={EventWarningType.INFO} />}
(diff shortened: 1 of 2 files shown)


src/components/EventCard/index.js
```diff
@@ -92,34 +92,36 @@ export default class EventCard extends Component {
...
-                <FavoriteButton eventAddress={address} />
-              </div>
-              <div className={classes.eventCardNameBundle}>
-                <div className={classes.eventCardNameFlex}>
-                  <div className={classes.eventCardInfoItem}>
-                    <FormattedMessage id='str.arbitrationReward' defaultMessage='Arbitration Reward' />{`: ${arbitrationRewardPercentage}%`}
-                  </div>
-                  <div className={classes.eventCardInfoItem}>
-                    <RaisedAmount amount={totalBets} />
+              <div className={classes.upper}>
```
dwalintukan: cant you just add the CSS to `eventCardSection`?

src/components/EventCard/styles.js
```diff
@@ -47,6 +53,14 @@ const styles = (theme) => ({
     fontSize: '1.05rem',
     textAlign: 'left',
     whiteSpace: 'nowrap',
+    [theme.breakpoints.down('xs')]: {
```
dwalintukan: you should be able to add the `noWrap` prop on Typography to handle ellipsizing it.
https://material-ui.com/api/typography/

dwalintukan: this is weird cause it already has ellipsis on the class. ya try using the material ui prop.

src/components/EventCard/index.js
henryzbw: we are taking all the stuff above options as one part, and options as another part, so I can add some space between these two parts, since on mobile size, it won't have height for the card.

src/components/EventCard/styles.js
henryzbw: Oh, its intentionally to show all the title and not ellipsis it. TT want when it's on mobile size(only one card per row) we show full title like first pic, if its on larger screen(2 or 3 cards per row), we ellipsis the title.

dwalintukan:
🚢

PR #1492, opened 2019-06-17
No pull

BernardHan:


src/scenes/Event/store.js
```diff
@@ -198,9 +203,10 @@ export default class EventStore {
     reaction(
       () => this.app.global.syncBlockNum,
       async () => {
-        if (this.app.ui.location === Routes.EVENT) {
+        if (this.eventNeedUpdate && this.app.ui.location === Routes.EVENT) {
           await this.initEvent();
         }
+        this.toggleEventNeedUpdate();
```
henryzbw: put it inside {}

src/stores/GlobalStore.js
```diff
@@ -38,6 +40,13 @@ export default class GlobalStore {
     return this.socketOnline && this.internetOnline;
   }
 
+  @computed get getBalanceNeedUpdate() {
```
henryzbw: is it used?

src/stores/UiStore.js
```diff
@@ -14,6 +14,7 @@ export default class UiStore {
   @observable currentTimeUnix = 0;
   @observable noWalletDialogVisible = false;
   counterInterval = null;
+  historyNeedUpdate = false; // for after create event, history gets init, and know need to pull for new create event tx
```
henryzbw: does it need to be observed?

Issue #1489, opened 2019-06-17
Remove all pulling data on new block


BernardHan:
1. event detail, only pull data on bet, set result, vote, withdraw
2. history page?

PR #1501, opened 2019-06-18
Async call

BernardHan:
#1489 

Instead of pulling data when the last txid is pending, I just pull all the pending list, if the list is not 0, then means it needs to keep pulling.

I still made the query compatible with using txid though. The reason I did not use it is that the mobile version currently still has the JS flow terminated problem, and it can never get the txid.


dwalintukan:
will review after the corresponding server PRs are addressed. logic might change here if the server does.

PR #1523, opened 2019-06-21
Test music

henryzbw:
TT wants to click logo 18 times to play music


src/components/NavBar/components/ImageLocaleWrapper/index.js
```diff
@@ -25,12 +25,14 @@ export default class ImageLocaleWrapper extends Component {
   }
 
   render() {
-    const { intl, src, alt, className } = this.props;
+    const { intl, src, alt, className, onClick } = this.props;
     return (
+      // eslint-disable-next-line
```
dwalintukan: what needs to be disabled?

src/components/NavBar/index.js
```diff
@@ -43,7 +56,11 @@ export default class NavBar extends Component {
         <Collapse in={!ui.searchBarMode}>
           <Toolbar className={classes.navBarWrapper}>
             <div className={classes.navSection}>
-              <BodhiLogo {...this.props} />
+              <BodhiLogo {...this.props} onClick={() => this.handleClickLogo()} />
+              {
+                // eslint-disable-next-line
```
dwalintukan: is the audio class not allowed?

src/scenes/App/index.js
```diff
@@ -28,6 +29,7 @@ const App = injectIntl(observer(({ classes, intl, match: { url }, store, store:
     {!store.loading && (
       <Fragment>
         <div className={classes.container}>
+          <NavBar />
```
dwalintukan: no props need to be passed to it?

src/components/NavBar/components/ImageLocaleWrapper/index.js
henryzbw: will have this warning, Non-interactive elements should not be assigned mouse or keyboard event listeners

src/components/NavBar/index.js
henryzbw: will have this error: Media elements such as <audio> and <video> must have a <track> for captions

src/scenes/App/index.js
henryzbw: don't think we need any props. all the page route is working

Issue #1539, opened 2019-07-02
Use EventStats and GlobalStats


dwalintukan:
- remove all calculateWinnings calls to blockchain
- use new EventStats in event detail
- use new GlobalStats in Global Leaderboard

PR #1546, opened 2019-07-10
Remove calculatewinnings

BernardHan:
#1539

### DIFF
--- a/src/network/graphql/queries.js
+++ b/src/network/graphql/queries.js
@@ -19,8 +19,6 @@ import {
   SYNC_INFO,
   TOTAL_RESULT_BETS,
   ALL_STATS,
-  PAGINATED_MOST_BETS,
-  PAGINATED_BIGGEST_WINNER,
   PAGINATED_LEADERBOARD_ENTRY,
 } from './schema';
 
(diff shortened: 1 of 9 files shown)


src/network/graphql/queries.js
```diff
@@ -404,21 +403,22 @@ export async function allStats(client) {
...
- * Queries the biggest winners based on an event address.
+ * Queries the global leaderboard entry based on a user address.
  * @param {ApolloClient} client Apollo Client instance.
  * @param {object} args Arguments for the query.
  * @return {object} Query result.
  */
-export async function biggestWinners(client, args) {
-  return new GraphQuery(client, QUERY_BIGGEST_WINNERS, args).execute();
+export async function globalLeaderboardEntries(client, args) {
+  const res = await new GraphQuery(client, QUERY_GLOBAL_LEADERBOARD_ENTRIES, args).execute();
```
dwalintukan: just return this. no reason to assign it first unless you were using `res`.

BernardHan: oh right.... this was for testing, forgot to change back.

PR #1551, opened 2019-07-17
Ans

BernardHan:
# Pull Request Template

## Checklist

- [x] Added CHANGELOG.md entry under Current section.
- [x] If I added new functionality, I added tests covering it.
- [x] If I fixed a bug, I added a regression test to prevent the bug from silently reappearing again.
- [x] I checked whether I should update the docs and did so if necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Current
 
 - [#1549] - Add Pull Request template and add CHANGELOG.md
+- [#1551] - Add Address Name Service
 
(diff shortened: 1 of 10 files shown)


henryzbw (approved):
LGTM

dwalintukan:
suggest to create a reusable component to handle this if it has a name or not.
```
<Address address={address} name={name} />

and in the comp:
if name is not undef/null/empty, show name, otherwise show address.
```

